Issue #7, opened 2021-09-06
Improve PostgrestFilterBuilder Semantics for work in Swift


chFlorian:
# Feature request

Right now the PostgrestFilterBuilder mirrors the Javascript API design. This might work great there, but it's not the kind of descriptive semantics that Swift users are used to. 

```swift
phfts(column: column, query: query, config: config)
```

Without knowing any context it is impossible to understand (and even then hard to read) what this expression does.

## Describe the solution you'd like

More descriptive function names would be great.

```swift
phrasetoFullTextSearch(column: String, query: String, config: String? = nil)
```

PR #8, opened 2021-09-06
Add PostgrestFilterBuilderSemantics

chFlorian:
## Issue
[Issue Link](https://github.com/supabase/supabase-swift/issues/7)

This PR introduces semantics updates for PostgrestFilterBuilder

```swift
phfts(column: column, query: query, config: config)
```

## What is the new behavior?

```swift
phrasetoFullTextSearch(column: String, query: String, config: String? = nil)
```

## Additional context

The idea is to improve semantics by providing descriptive function names. These are implemented as an extension to PostgrestFilterBuilder, which allows to keep the existing function names (aka the mirrors of the JS library).

### DIFF
--- a/Sources/Supabase/PostgrestSemantics/PostgrestFilterBuilderSemantics.swift
+++ b/Sources/Supabase/PostgrestSemantics/PostgrestFilterBuilderSemantics.swift
@@ -1,0 +1,59 @@
+import PostgREST
+
+extension PostgrestFilterBuilder {
+    public func equals(column: String, value: String) -> PostgrestFilterBuilder {
+        eq(column: column, value: value)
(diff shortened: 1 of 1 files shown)


chFlorian:
@satishbabariya should this be moved to supabase/postgrest-swift?

grsouza:
Hey @chFlorian thanks for the contribution, and yeah I think this change fits more in the `supabase/postgrest-swift` repo.

Other question, by following the convention, the case wouldn't be `phraseToFullTextSearch` with upper `T` after the `phrase`?

chFlorian:
Alright, then I will move this PR over to the other repo and delete it here once done.

PR #11, opened 2021-09-17
README update & added "platforms" in Package.swift 

AngCosmin:
Hey 👋

I updated the README to match the current implementation and I added the `platforms` field in `Package.swift` because without it I couldn't import `Supabase` inside the iOS project.

Have a nice day ☀️,
Cosmin

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -5,17 +5,17 @@ import PackageDescription
 
 let package = Package(
     name: "Supabase",
-    products: [
-        .library(
-            name: "Supabase",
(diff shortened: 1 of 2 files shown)


grsouza (commented):
Thanks @AngCosmin 

I didn't understand the changes on the example from README, you removed the do/catch block?

The other things are great.

AngCosmin:
Hey @grsouza 

I removed the do/catch blocks because they are not needed. No exception is thrown

image

Issue #52, opened 2023-01-02
Fails to build due to PostgREST master branch dependency


glowcap:
# Bug report

## Describe the bug

Build attempts fail due to changes in PostgREST master branch

- `FunctionsClient` init change
- `PostgrestClient` init change
- `FunctionsHTTPClient` removal
- `PostgrestHTTPClient` removal

## To Reproduce

1. Create new project (or have existing project)
2. Add supabase-swift SPM dependency as described in README
3. Attempt to build the project
4. See errors in `SupabaseClient.swift`

## Expected behavior

Should build without issue

## Screenshots

<img width="233" alt="Screenshot 2023-01-02 at 1 00 27 PM" src="https://user-images.githubusercontent.com/10408147/210270103-3692ea91-058c-4f5f-94a9-13b554c360c5.png">

## Additional context

Switching to the Release Candidate #45 most recent commit solves the issue (albeit not ideal)

`.package(name: "Supabase", url: "https://github.com/supabase/supabase-swift.git", branch: "release-candidate")`

It seems like shifting the master branch to tagged dependencies from here on would avoid issues like these popping up

glowcap:
Created [PR](https://github.com/supabase-community/supabase-swift/pull/53) to resolve the issue

grsouza:
PR merged and new release created with the following fix, https://github.com/supabase-community/supabase-swift/releases/tag/0.0.12

Thanks @glowcap .

---

Would be very nice if you used the release-candidate branch on your project, for helping catching any bugs on that.

PR #45, opened 2022-10-28
Release Candidate

grsouza:
Depends on:
- https://github.com/supabase-community/postgrest-swift/pull/34
- https://github.com/supabase-community/gotrue-swift/pull/37

PR #53, opened 2023-01-03
Fixed buid errors - #52

glowcap:

Fixes build errors mentioned in issue #52 by:

- Updating Package file
- Updating `SupabaseClient.swift`

Fails to build due to breaking changes in `postgrest-swift`

## What is the new behavior?

Now builds and runs as expected

## Additional context

Glad to help!

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.5
+// swift-tools-version:5.7
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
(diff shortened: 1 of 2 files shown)


grsouza:
Hy @glowcap thanks for this.

But the #45 will already solve this, I just need a bit more testing for merging the RC.

glowcap:
Hi @grsouza.  I understand, but your PR goes way beyond the scope of the fix and will probably take more time to review. In the meantime, people like myself can't reset package caches or run any SPM updates without running into this issue.

This simply let's people move forward with their work while you wrap up your RC branch and allow it to be properly reviewed and merged

grsouza (approved):

PR #61, opened 2023-02-18
Update ReadMe with Insert and Select Queries

maail:
Added in Insert and Select Queries for Supabase Swift

Added documentation for Insert and Select Queries

Current documentation does not show basic CRUD functionalities. It's only available as tests on Postgrest-Swift repo, which not everyone might find easily.

## What is the new behavior?

Added documentation for Insert and Select Queries

## Additional context

I'll add in Update and Delete functionality once I've used them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,14 +198,12 @@ let query = client.database
             .insert(values: insertData,
                     returning: .representation) // you will need to add this to return the added data
             .select(columns: "id") // specifiy which column names to be returned. Leave it empty for all columns
+            .single() // specify you want to return a single value.
             
 Task {
     do {
-        if let response = try? await query.execute(),
-           let data = try? response.underlyingResponse.data  {
-           let returnArray = try JSONDecoder().decode([InsertModel].self, from:  data)
-            print("### Returned: \(returnArray.first)")
-        }
+        let response: InsertModel = try await query.execute().value
+        print("### Returned: \(response)")
     } catch {
         print("### Insert Error: \(error)")
     }
--- a/README.md
+++ b/README.md
@@ -219,15 +219,13 @@ let insertData = InsertModel(title: "Test", desc: "Test Desc")
 let query = client.database
             .from("{ Your Table Name }")
             .select() // keep it empty for all, else specify returned data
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -173,6 +173,69 @@ Task {
...
+    do {
+        if let response = try? await query.execute(),
+           let data = try? response.underlyingResponse.data  {
+           let returnArray = try JSONDecoder().decode([InsertModel].self, from:  data)
+            print("### Returned: \(returnArray.first)")
+        }
+    } catch {
+        print("### Insert Error: \(error)")
+    }
+}
```
grsouza, line 212: You don't need to manually decode the response, every `Respnse` returned has ` value` property that returns the decoded model, you just need to help Swift figure it out the type.

Also when creating the query, you can specify that you want a `single` value to be returned.

```suggestion
let query = client.database
            .from("{ Your Table Name }")
            .insert(values: insertData,
                    returning: .representation) // you will need to add this to return the added data
            .select(columns: "id") // specifiy which column names to be returned. Leave it empty for all columns
            .single() // specify you want to return a single value.
            
Task {
    do {
        let response: InsertModel = try await query.execute().value
        print("### Returned: \(response)")
    } catch {
        print("### Insert Error: \(error)")
    }
}
```

README.md
```diff
@@ -173,6 +173,69 @@ Task {
...
+    do {
+        if let response = try? await query.execute(),
+           let data = try? response.underlyingResponse.data  {
+            let returnArray = try JSONDecoder().decode([InsertModel].self, from:  data)
+            print("### Returned: \(returnArray.first)")
+        }
+    } catch {
+        print("### Insert Error: \(error)")
+    }
+}
```
grsouza, line 236: Same thing as the insert example, you can define as `single` and use `value`'s response to fetch a decoded object.

grsouza (changes requested):

README.md
maail: Oh that's pretty cool.

README.md
maail: @grsouza Updated.

grsouza (approved):

PR #65, opened 2023-02-27
Fix small error in README Social Login

iamlouislab:

Docs update.

The current example provided in the README is incorrect. 

## What is the new behavior?

The change fixes the example provided

## Additional context

This is a similar change that was made in the last pull requests but it was not taken into consideration on the social login


maail:
@iamlouislab looks good. Thanks Louis!

PR #184, opened 2023-12-10
Add Basic Windows & Linux Support

brianmichel:

This adds the ability to use this SDK on Windows and Linux environments.

Currently, the SDK doesn't compile on Linux or Windows.

## What is the new behavior?

This PR adds support to compile, test, and run the SDK on Windows and Linux.

## Additional context

As with all things, Swift on non-Darwin platforms comes with some caveats:
- The keychain doesn't exist on non-Darwin platforms. I've made a stub to get the SDK working, and left comments on what will probably need to happen in a different PR (tl;dr it will probably be easier to add the Windows support here compared to Linux since there are stable APIs on Windows to use)
- All of the test files had to be broken out on different platforms for two reasons:
  1. Windows uses crlf line endings instead of Darwin and Linux which use lf line endings, this is quite annoying, but that's life
  2. Foundation on non-Darwin behaves differently and does some auto-canonicalization of http headers (i.e. `apikey` becomes `Apikey`), except anything starting with an `X-` which creates ordering issues when using the textual format.

I was unable to get the examples to run on my mac since it looks like they required a different team to get setup, so I'm hoping that running them on this repo's CI will work, but I'm seeing some success on my fork https://github.com/brianmichel/supabase-swift/actions/runs/7160617987/job/19495202392?pr=1

Looking forward to the discussion :D


Sources/GoTrue/Internal/PKCE.swift
```diff
@@ -1,19 +1,20 @@
-import CryptoKit
+import Crypto
```
brianmichel, line 1: It seemed more straightforward to have one dependency on Crypto which is open source and designed to be portable across operating systems.

Sources/_Helpers/URLSession+AsyncAwait.swift
```diff
@@ -0,0 +1,43 @@
+#if canImport(FoundationNetworking)
+import Foundation
+import FoundationNetworking
+
+public enum URLSessionPolyfillError: Error {
+  case noDataNoErrorReturned
+}
+
+extension URLSession {
```
brianmichel, line 9: Foundation on non-Darwin platforms doesn't have async/await support, so these are just shims for those APIs

Sources/_Helpers/URLSession+AsyncAwait.swift
```diff
@@ -0,0 +1,43 @@
...
+        if let error {
+          continuation.resume(throwing: error)
+        } else if let data, let response {
+          continuation.resume(returning: (data, response))
+        } else {
+          continuation.resume(throwing: URLSessionPolyfillError.noDataNoErrorReturned)
+        }
+      })
+
+      task.resume()
```
grdsdev, line 22: Please add support for canceling the data task, when the Task is canceled.

Sources/_Helpers/URLSession+AsyncAwait.swift
```diff
@@ -0,0 +1,43 @@
...
```
grdsdev, line 38: Please add support for canceling the upload task, when the Task is canceled.

Tests/GoTrueTests/GoTrueClientTests.swift
```diff
@@ -11,7 +11,15 @@ import ConcurrencyExtras
...
 
+#if canImport(FoundationNetworking)
+import FoundationNetworking
+#endif
+
 final class GoTrueClientTests: XCTestCase {
+  fileprivate var api: APIClient!
+
+  #if !os(Linux)
+  // `fullfillment` doesn't exist on Linux yet.
```
grdsdev, line 22: Please use [`wait(for:timeout:)`](https://developer.apple.com/documentation/xctest/xctestcase/2806856-wait) instead of fulfillment then.

Tests/GoTrueTests/RequestsTests.swift
```diff
@@ -160,10 +164,14 @@ final class RequestsTests: XCTestCase {
   }
 
   func testSessionFromURL() async throws {
+    // For some reason this crashes the testing bundle
+    // on Linux, macOS, and Windows, skipping it.
+    try XCTSkipIf(true)
```
grdsdev, line 169: Is there any log regarding the crash? At least on macOS, it should work.

Sources/GoTrue/GoTrueLocalStorage.swift
```diff
@@ -30,3 +32,37 @@ struct KeychainLocalStorage: GoTrueLocalStorage {
...
   }
 }
+#else
+// Secure storage in Linux seems to be quite different than say macOS and Windows
+// The keyctl API set exists, but it doesn't persist across reboots like the
+// keychain on macOS and the password vault (wincred.h) on Windows does. So this
+// provides a bit of a challenge of how to implement this on Linux.
+//
+// Realistically, you probably want to pull in gpg and build your own implementation.
+final class KeychainLocalStorage: GoTrueLocalStorage {
```
grdsdev, line 42: I'd prefer to force the developer to pass a `GoTrueLocalStorage` implementation on client initialization in case of a non-darwin platform instead of providing a default implementation.

Sources/GoTrue/Internal/PKCE.swift
grdsdev: LGTM, thank you.

Sources/GoTrue/Internal/FixedWidthInteger+Random.swift
```diff
@@ -0,0 +1,30 @@
+import Foundation
+
+// Borrowed from the Vapor project, https://github.com/vapor/vapor/blob/main/Sources/Vapor/Utilities/Array%2BRandom.swift#L14
+extension FixedWidthInteger {
```
grdsdev, line 4: Lets keep all this `internal`.

Sources/_Helpers/AssertSnapshot+Platform.swift
```diff
@@ -0,0 +1,50 @@
+import Foundation
+import SnapshotTesting
+import XCTest
+
+/// A platform aware version of the standard `assertSnapshot`
+/// This is used to point to different subdirectories of the `__Snapshots__`
+/// directory. On different platforms, the different textual representations
+/// are different for a number of reasons, so this gives us a way to differentiate
+/// where needed.
+public func platformSpecificAssertSnapshot<Value, Format>(
```
grdsdev, line 10: I understood the issue with the default implementation on the `assertSnapshot` method, but I'd prefer to open a discussion on the `swift-snapshot-testing` repository to check if there is any other solution to this.

grdsdev (changes requested):
Thank you for the hard work on PR this, it looks awesome.

I left a few comments for discussion.

Thanks again!

brianmichel:
Hey @grdsdev sorry for the delay, I'm going to pick this back up today and see if I can address some of the changes you've requested.

Tests/GoTrueTests/GoTrueClientTests.swift
brianmichel: Looking at the sources again, I can't see where Linux doesn't actually have this, let me try to re-run this test and see what happens...

brianmichel: So while I couldn't use `fulfillment` (doesn't exist on Linux), and couldn't use `wait` (since we're in an async context), I was able to use a simple semaphore to accomplish the same thing. Let's see how that does across platforms.

brianmichel: Looks like that works well!

brianmichel:
Hey @grdsdev I'm seeing errors regarding swift evolution support, which trace back to snapshot testing and I see an open issue on their repo https://github.com/pointfreeco/swift-snapshot-testing/issues/536 has this CI job worked before?

brianmichel:
Hey @grdsdev I'm pretty close to responding to all of your feedback, let me know what you think of the approach I've uploaded, looks like all of CI that was passing is passing again https://github.com/brianmichel/supabase-swift/actions/runs/7240544944/job/19723650563?pr=1. I'll open up convos with the Pointfree folks in a little bit. 

I might have to rebase this another day as I imagine it's going to be quite messy.

Sources/GoTrue/Storage/WinCredLocalStorage.swift
```diff
@@ -0,0 +1,82 @@
...
+
+    if !CredDeleteW(targetName, credentialType, 0) {
+      let lastError = GetLastError()
+      debugPrint("Unable to remove key from credential vault, got error code \(lastError)")
+
+      throw WinCredLocalStorageError.windows(lastError)
+    }
+  }
+}
+#endif
```
brianmichel, line 82: Forgot to add new line, will add.

Sources/GoTrue/Storage/LocalStorageEngines.swift
```diff
@@ -0,0 +1,15 @@
...
+    KeychainLocalStorage(service: "supabase.gotrue.swift", accessGroup: nil)
+  }()
+  #endif
+
+  #if os(Windows)
+  public static let wincred: some GoTrueLocalStorage = {
+    WinCredLocalStorage(service: "supabase.gotrue.swift")
+  }()
+  #endif
+}
```
brianmichel, line 15: Forgot to add new line, will add.

Sources/_Helpers/AssertSnapshot+Platform.swift
brianmichel: Issue re-opened on the snapshot testing repo https://github.com/pointfreeco/swift-snapshot-testing/issues/815

brianmichel: I ended up fixing this so it's no longer needed!

brianmichel:
Hey @grdsdev I believe I've fixed all of the issues and my latest CI run was all green https://github.com/brianmichel/supabase-swift/pull/1/commits/2d8c882e1f4193490c39a7c62bf988b10e9447b7 there are likely some lingering questions that should be addressed, but I'd say this is 100% ready for a re-review and potentially ready for merging.

Examples/Examples/ExamplesApp.swift
```diff
@@ -19,6 +19,6 @@ struct ExamplesApp: App {
 }
 
 let supabase = SupabaseClient(
-  supabaseURL: Secrets.supabaseURL,
+  supabaseURL: Secrets.supabaseURL.absoluteString,
```
brianmichel, line 22: With the new convenience methods we ended up with everything being a string in the interface, would you like to keep this as a URL? If so I can edit the tests (I think they are passing in strings)

Sources/Auth/AuthClient.swift
```diff
@@ -34,7 +34,7 @@ public actor AuthClient {
       url: URL,
       headers: [String: String] = [:],
       flowType: AuthFlowType = Configuration.defaultFlowType,
-      localStorage: AuthLocalStorage = Configuration.defaultLocalStorage,
+      localStorage: AuthLocalStorage,
```
brianmichel, line 37: We can no longer guarantee a default since there is no good way to provide a standard implementation in Linux (there is no similar Keychain/WinCred API)

Tests/AuthTests/GoTrueClientTests.swift
```diff
@@ -31,11 +39,11 @@ final class AuthClientTests: XCTestCase {
             $0.append(event)
           }
 
-          expectation.fulfill()
+          semaphore.signal()
```
brianmichel, line 42: Using a semaphore since it was the best cross-platform tool to wait for some async changes

Tests/AuthTests/RequestsTests.swift
```diff
@@ -395,7 +403,7 @@ final class RequestsTests: XCTestCase {
 
     let configuration = AuthClient.Configuration(
       url: clientURL,
-      headers: ["apikey": "dummy.api.key", "X-Client-Info": "gotrue-swift/x.y.z"],
+      headers: ["Apikey": "dummy.api.key", "X-Client-Info": "gotrue-swift/x.y.z"],
```
brianmichel, line 406: I ended up changing all of the references to `apikey` in the tests to `Apikey` since there is some annoying casing differences generated in the snapshots across platform. This felt acceptable as HTTP headers are specified to be case-insensitive and this was just a quirk of the tests.

Examples/Examples/ExamplesApp.swift
grdsdev: Let's keep as a URL

grdsdev (changes requested):

Sources/_Helpers/URLSession+AsyncAwait.swift
```diff
@@ -0,0 +1,116 @@
...
+          } else if let data, let response {
+            continuation.resume(returning: (data, response))
+          } else {
+            continuation.resume(throwing: URLSessionPolyfillError.noDataNoErrorReturned)
+          }
+        })
+
+        helper.register(task)
+
+        task.resume()
```
jamieQ, line 83: i _think_ this has no guaranteed ordering relative to the effective `helper` calls. i.e. the task will start running ~here even if the on-actor `actuallyCancel`/`actuallyRegister` calls haven't yet executed. does that matter?

Examples/Examples/ExamplesApp.swift
brianmichel: I ended up removing the convenience initializers since their signatures conflicted with the default initializer, but also I feel like it's better to force people to specify this auth option on all platforms since we can't guess what people are building exactly. 

It's entirely possible to be writing code on macOS for a codebase that's going to be used on Linux. This would cause confusion for consumers of this package making the assumption that since they didn't make a specific choice about the secure storage engine. 

I realize this is quite opinionated, and I can probably provide some static functions or something like that to configure 'default' clients, but it felt worth checking this assumption since it can easily create confusion in a cross platform codebase.

Sources/_Helpers/URLSession+AsyncAwait.swift
```diff
@@ -0,0 +1,116 @@
...
+      await actuallyCancel()
+    }
+  }
+
+  private func actuallyRegister(_ task: URLSessionTask) {
+    switch state {
+    case .registered:
+      preconditionFailure("Attempting to register another task while the current helper already has a registered task!")
+    case .cancelled:
+      preconditionFailure("Attempting to register a task while the current handler is in the cancelled state!")
```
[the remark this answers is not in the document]
brianmichel: Ah yeah you're right this was an oversight by me!

Sources/_Helpers/URLSession+AsyncAwait.swift
```diff
@@ -0,0 +1,116 @@
...
+    case .initialized:
+      state = .registered(task)
+
+    }
+  }
+
+  private func actuallyCancel() {
```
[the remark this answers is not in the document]
brianmichel: Yeah, you're right this should set itself to cancelled since that could absolutely be called first and any subsequent registrations should be avoided and the passed in task should be cancelled since we're already in the cancelled state.

Sources/_Helpers/URLSession+AsyncAwait.swift
brianmichel: Yeah I had this same question when looking at the reference code. I'm not sure if it's totally avoidable, and we might end up with whatever quirks the reference has as well. 

One complication is that the `onCancel` closure is required to be synchronous so I'm not sure how to improve upon what the async http client folks have already cooked up. Any ideas?

Sources/_Helpers/URLSession+AsyncAwait.swift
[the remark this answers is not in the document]
brianmichel: Fixed

Sources/_Helpers/URLSession+AsyncAwait.swift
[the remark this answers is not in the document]
brianmichel: Fixed

Sources/_Helpers/URLSession+AsyncAwait.swift
```diff
@@ -0,0 +1,121 @@
...
+    state = .cancelled
+
+    switch state {
```
jamieQ, line 59: note: this will now always hit the `.cancelled` case due to the location of the assignment above. the state change should probably occur after reading the existing state.

brianmichel: ah yeah, my bad, fixed

Sources/_Helpers/URLSession+AsyncAwait.swift
jamieQ: i think one possibility would be to merge resumption of the task into the `register` invocation (`registerAndResume(...)`?). i'm not sure there would be any issues in practice with the code as-is (cancellation now seems like it will eventually be propagated), but since swift concurrency gives no assurances about the ordering of submitted tasks, or how long it might take before they're run, it seems like there's a theoretical opening for weird behavior where the data task runs for a non-trivial amount of time before the cancel/register calls are made (or could it even run to completion before registering/canceling in some edge case?). the main benefit of moving resumption into the registration logic might be to just eliminate having to think through this sort of scenario.

again, i don't think a change is necessarily needed here... mostly this is just a question out of curiosity.

grdsdev (approved):
Awesome work, I'll have it merged soon, thank you!

Issue #177, opened 2023-12-06
Investigate possible issue with `insert` method


grdsdev:
# Bug report

## Describe the bug

There is a possible issue with the `insert` method reported in the following tweet https://twitter.com/Archetapp/status/1731997988288147553

adhanson6:
I had seen this exact error before when I didn't have the 'returning: .representation' as a parameter to the insert method.

grdsdev:
Closing as it hasn't been updated for a while.

If someone faces this issue again, feel free to re-open it.

PR #204, opened 2023-12-20
Fix issue where nil default returning parameter caused bug in insert …

bryan-vh:

#177 

Potential bug referenced here: https://twitter.com/Archetapp/status/1731997988288147553

## What is the new behavior?

No longer results in bug due to default parameter being set to `.representation`

## Additional context

N/A


grdsdev:
Hi @bryan-vh  thanks for the fix.

Could you first add a code example that triggers the error? And that gets fixed by this change?

PR #214, opened 2023-12-29
refactor: Realtime V2

grdsdev:
## Changes

Adds a new implementation of Realtime that makes the connection more reliable by adopting modern Swift constructs such as actors, async/await, and AsyncSequence.


mkaulfers:
Looking forward to this one. Sheesh updating from `0.3.0` is absolutely atrocious.

PR #250, opened 2024-02-27
feat: Add optional "referencedTable" parameter to OR PostGREST filter

mpross512:

Feature

When using the or filter, the columns have to be in the table referenced in the "from" clause

## What is the new behavior?

Users can now specify an optional "referencedTable" parameter to the or filter, allowing the or to reference other tables besides the main table in the "from" clause.

## Additional context

image

### DIFF
--- a/Sources/PostgREST/PostgrestFilterBuilder.swift
+++ b/Sources/PostgREST/PostgrestFilterBuilder.swift
@@ -22,9 +22,10 @@ public class PostgrestFilterBuilder: PostgrestTransformBuilder {
     return self
   }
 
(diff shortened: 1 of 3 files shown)


mpross512:
You can find the PostGREST documentation on embedded filters [here](https://postgrest.org/en/latest/references/api/resource_embedding.html#embedded-filters)

grdsdev (commented):
Thanks for the contribution.

Can tou please add a test case for the or method call with a referenced table in https://github.com/supabase-community/supabase-swift/blob/main/Tests/PostgRESTTests/BuildURLRequestTests.swift

Thank you.

grdsdev (approved):

PR #293, opened 2024-03-30
Fixes #292

### LINKED ISSUE #292: Broke `StoredSession` JSON backward-compatibility
# Bug report
I noticed, that after upgrading to 2.5.0, I was logged out with the following error while retrieving the existing session:

```
keyNotFound(CodingKeys(stringValue: "identityId", intValue: nil), Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "session", intValue: nil), CodingKeys(stringValue: "user", intValue: nil), CodingKeys(stringValue: "identities", intValue: nil), _JSONKey(stringValue: "Index 0", intValue: 0)], debugDescription: "No value associated with key CodingKeys(stringValue: \"identityId\", intValue: nil) (\"identityId\").", underlyingError: nil))
```

After digging through recent changes, I believe the changes in #274 have caused this issue. A new value `identityId` was added to the `UserIdentifiy` without providing a fallback for stored sessions.

## Describe the bug

The user gets logged out with the error above in the logs.

## To Reproduce
- Log in while using an older version 
- Upgrade to the latest release
- Run app again
-> User is logged out

## Expected behavior

There is some sort of backwards-compatibility between releases.

## System information

- OS: iOS
- Version of supabase-swift: 2.5.0

leoMehlig:

This PR fixes the bug described in #292.

1. Adds a new test class `StoredSessionTests`, which includes tests to decode the previous JSON and the current.
2. Adds a default value to `identityId` to make it backwards-compatible with the `StoredSession` in 2.4.0

Decoding an old stored session would fail.

## What is the new behavior?

Decoding an old stored session uses a default (all zeros) value for `identityId`

## Additional context

An alternative approach would be to make `identityId` optional, but that requires changes elsewhere in the SDK. The upside of that approach is, that it is more explicit and would not require custom Codable code.

### DIFF
--- a/Sources/Auth/Types.swift
+++ b/Sources/Auth/Types.swift
@@ -220,6 +220,44 @@ public struct UserIdentity: Codable, Hashable, Identifiable, Sendable {
     self.lastSignInAt = lastSignInAt
     self.updatedAt = updatedAt
   }
+  
+  enum CodingKeys: CodingKey {
+    case id
+    case identityId
+    case userId
+    case identityData
+    case provider
+    case createdAt
+    case lastSignInAt
+    case updatedAt
+  }
+  
+  public init(from decoder: any Decoder) throws {
+    let container: KeyedDecodingContainer<UserIdentity.CodingKeys> = try decoder.container(keyedBy: UserIdentity.CodingKeys.self)
+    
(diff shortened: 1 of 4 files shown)


grdsdev:
Thanks @leoMehlig for the PR, I made a few changes to it, going to push it.

PR #295, opened 2024-04-01
test(auth): add tests for auto verifying stored session compatibility

grdsdev:

Bug fix, feature, docs update, ...

Please link any relevant issues here.

## What is the new behavior?

Feel free to include screenshots if it includes visual changes.

## Additional context


leoMehlig:
Quick question: How is the test now verifying the backwards comp to 2.4.0? The JSON file represents the 2.5.0 schema, right?

grdsdev:
Hey, it doesn't verify specifically the 2.4.0 compatibility, but now it stores values in a JSON file, so the test verifies it can load the previous stored value (asserting that it is compatible with older versions), and then update the stored value.

If we make incompatible change to `StoredSession` type, it will fail to decode from the current `local-storage.json` file.

PR #299, opened 2024-04-01
feat(auth): add `signInWithOAuth`

grdsdev:

Close https://github.com/supabase-community/supabase-swift/issues/257

## How OAuth worked before?
Before the OAuth implementation consisted of 3 steps:

1. Call the `getOAuthSignInURL` method that returned a `URL` type.
2. Use the URL returned in step 1 to initiate the OAuth flow, this was up to the developer to implement, on completion of this step, it should have a result URL that contains the `token/code`.
3. Use the result URL from step 2 to call `session(from:)` method that would parse the URL and import the session into the client.

## How it works after this PR?

Everything from  **How OAuth worked before?** is still valid in case user wants full control of the flow.

But now there are 2 convenience methods (built on top of the already existent methods).

### Sign in leveraging the steps mentioned above but injecting the **Step 2**

```swift
let session = try await signInWithProvider(provider: .google) { url in
    let resultURL = await customFlowImplementation(url: url) // Use url for initiating the flow, some options could be implementing a web view, using `openURL` method from `UIApplication`, and manually observing for deep links.
    return resultURL // Containing #token=abc.def.ghi
}
```

### Sign in using ASWebAuthenticationSession on the platforms that supports it (Recommended)

```swift
let session = try await signInWithProvider(provider: .google) { session in 
    // additional configuration to session object
    session.presentationContextProvider = YourCustomImplementation()
}
```

### Other changes introduced

- Instead of providing a `redirectURL` on all methods, you can pass one when initializing the client, and it will be used as the default if none is provided in the method.


dshukertjr (approved):
Just to double-check, the following code on the PR body is actually meant to be `signInWithOAuth`, correct?

>let session = try await signInWithProvider(provider: .google) { session in 
    // additional configuration to session object
    session.presentationContextProvider = YourCustomImplementation()
}

grdsdev:
Yes @dshukertjr thanks for that.

PR #319, opened 2024-04-02
chore: automate release with release-please

grdsdev:


.github/workflows/release.yml
```diff
@@ -0,0 +1,18 @@
+on:
+  push:
+    branches:
+      - chore/release-please
```
grdsdev, line 4: remember to revert to `main`

PR #325, opened 2024-04-05
fix(postgrest): race condition when setting fetchOptions and execute method call

grdsdev:

Fixes a bug which could lead to a race condition when executing the request on Postgrest.

Lock was being acquired first for updating the fetchOptions and then inside the `execute` method for executing the request, this could lead to the `execute` method to use the wrong fetch options.

## What is the new behavior?

Move assignment of `fetchOptions` inside `execute` method, on a single lock acquisition.

### DIFF
--- a/Sources/PostgREST/PostgrestBuilder.swift
+++ b/Sources/PostgREST/PostgrestBuilder.swift
@@ -51,11 +51,7 @@ public class PostgrestBuilder: @unchecked Sendable {
   public func execute(
     options: FetchOptions = FetchOptions()
   ) async throws -> PostgrestResponse<Void> {
-    mutableState.withValue {
-      $0.fetchOptions = options
-    }
-
-    return try await execute { _ in () }
+    try await execute(options: options) { _ in () }
   }
 
   /// Executes the request and returns a response of the specified type.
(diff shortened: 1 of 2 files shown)


steve-chavez:
> Lock was being acquired first for updating the fetchOptions and then inside the execute method for executing the request, this could lead to the execute method to use the wrong fetch options.

@grdsdev Q: Why are locks needed on the client? Can't it be stateless? (Since PostgREST requests are stateless)

grdsdev:
@steve-chavez 

Check the following test:

```swift
func testRaceCondition() async {
    let select = client.from("users").select()

    Task {
      print("Task 1")
      try await select.execute(options: FetchOptions(head: true))
    }

    Task {
      print("Task 2")
      try await select.execute(options: FetchOptions(head: false))
    }

    try? await Task.sleep(nanoseconds: NSEC_PER_SEC * 2)
  }
```

Task 1 and Task 2 can run on different threads, Task 2 can even start running before Task 1, that is all up to the Swift concurrency system.

steve-chavez:
@grdsdev Got it. Thanks for the explanation :+1:

PR #337, opened 2024-04-12
feat: remove support for Windows/Linux

grdsdev:

For the official support we're focusing only on the Apple platform, so I'm removing the current code handling Windows/Linux support.


mergesort:
Hey @grdsdev, I was wondering what suggestions you would have for me to continue using Supabase on Linux? I have a good amount of code using Supabase on Linux since it was added to the official library, so I consider this a really big breaking change not to be taken lightly.

brianmichel:
Personally I can understand the complexity of not wanting to support Windows as it's very much evolving, however I would say that removing all Linux support feels like a miss for me.

Why limit the usability and reach of an SDK, is this just a resourcing problem?

dshukertjr (approved):

grdsdev:
@mergesort and @brianmichel, thanks for bringing in your points to the discussion.

I decided to keep the current code as it is working, but we're not officially supporting Linux and Windows for now on.

I'll close this PR in favor of https://github.com/supabase/supabase-swift/pull/350 which fixes a couple of issues with Linux build and adds a **Supported Platform** to the README that makes this more clear.

mergesort:
Thank you so much @grdsdev, I truly appreciate that you're keeping unofficial support for Linux around! I'll make sure to do my best to fix any issues I encounter on Linux, and if not will create issues to see if we can work to solve them.

PR #345, opened 2024-04-18
feat(realtime): add closure based methods

grdsdev:

Add closure based methods for RealtimeChannel:

## Current behavior
Library currently exposes only methods that returns an `AsyncStream` to be used in Swift Concurrency model.

## New behavior
This PR add closure based methods `onPostgresChange`, `onBroadcast`, and `onPresenceChange`


dshukertjr (approved):
Nice!

Issue #252, opened 2024-02-28
Ability to stream function responses


tmcgann:
# Feature request

## Is your feature request related to a problem? Please describe.

Want to stream function responses. Use case: stream chat completions from AI provider.

## Describe the solution you'd like

Would love to augment the `FunctionsClient` to be able to stream data from a function response. Rough example of solution:

```swift
/// Implement the StreamResponseHandler protocol to handle the streamed data.
/// You could append received data chunks to a buffer, process data as it arrives, 
/// or update the UI (updates would be dispatched on the main thread).
protocol StreamResponseHandler {
    func didReceive(data: Data)
    func didCompleteWithError(_ error: Error?)
}

class StreamDataTaskHandler: NSObject, URLSessionDataDelegate {
    let streamHandler: StreamResponseHandler

    init(streamHandler: StreamResponseHandler) {
        self.streamHandler = streamHandler
    }

    func urlSession(_ session: URLSession, dataTask: URLSessionDataTask, didReceive data: Data) {
        // Forward the received data to the stream handler
        streamHandler.didReceive(data: data)
    }

    func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?) {
        // Notify the stream handler about completion or error
        streamHandler.didCompleteWithError(error)
    }
}

extension FunctionsClient {
    public func invokeWithStream(
        _ functionName: String,
        options: FunctionInvokeOptions = .init(),
        streamHandler: StreamResponseHandler
    ) async throws {
        // Setup URLSession with StreamDataTaskHandler as delegate
        let delegate = StreamDataTaskHandler(streamHandler: streamHandler)
        let session = URLSession(configuration: .default, delegate: delegate, delegateQueue: nil)
        
        // Rest of the setup similar to `rawInvoke`, but use the session with delegate
        let url = self.url.appendingPathComponent(functionName)
        var urlRequest = URLRequest(url: url)
        urlRequest.allHTTPHeaderFields = options.headers.merging(headers) { invoke, _ in invoke }
        urlRequest.httpMethod = (options.method ?? .post).rawValue
        urlRequest.httpBody = options.body
        
        // Start the data task
        let task = session.dataTask(with: urlRequest)
        task.resume()
        
        // Ensure to keep the session and delegate alive as long as needed
        // This may require storing references in the FunctionsClient or managing lifecycle appropriately
    }
}

```

## Describe alternatives you've considered

Make the request to the function outside of supabase-swift.

## Additional context

[...]

PR #346, opened 2024-04-18
feat(functions): add experimental invoke with streamed responses

grdsdev:

Close https://github.com/supabase/supabase-swift/issues/252

Add experimental method to invoke function with streamed responses

## Why is it sill experimental?

To implement the streamed responses, we need to set up a custom `URLSession` for setting its `delegate` property, which means that it doesn't use the `customFetch` that is provided to the `functions` package and all other `sub-packages`.

I prefer keeping it `Experimental` until we try to refactor the code so that all requests uses the same `URLSession` instance.

### DIFF
--- a/Sources/Functions/FunctionsClient.swift
+++ b/Sources/Functions/FunctionsClient.swift
@@ -60,16 +60,10 @@ public actor FunctionsClient {
     url: URL,
     headers: [String: String] = [:],
     region: FunctionRegion? = nil,
-	logger: (any SupabaseLogger)? = nil,
+    logger: (any SupabaseLogger)? = nil,
     fetch: @escaping FetchHandler = { try await URLSession.shared.data(for: $0) }
   ) {
-    self.url = url
-    self.headers = headers
-    if headers["X-Client-Info"] == nil {
-      self.headers["X-Client-Info"] = "functions-swift/\(version)"
-    }
-    self.region = region?.rawValue
-	http = HTTPClient(logger: logger, fetchHandler: fetch)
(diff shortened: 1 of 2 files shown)


dshukertjr:
So this PR introduces a new public method `_invoke()` to handle stream response, correct? Would it be better to rename it to something like `invokeStreamResponse`, or something along that line maybe?

grdsdev:
The `_` prefix is commonly used in Swift to signal an `experimental` or `private` method, but I'll rename it to `_invokeWithStreamedResponse` or something like that.

dshukertjr:
Didn't know that! Thanks for educating me! `_invokeWithStreamedResponse` sounds good!

dshukertjr (approved):

Sources/Functions/FunctionsClient.swift
```diff
@@ -130,33 +128,101 @@ public actor FunctionsClient {
   private func rawInvoke(
     functionName: String,
     invokeOptions: FunctionInvokeOptions
-  ) async throws -> (Data, HTTPURLResponse) {
+  ) async throws -> Response {
+    var request = Request(
+      path: functionName,
+      method: .post,
```
jareddr, line 134: This line is causing a major bug in the library.  Upon upgrading our client version, we can no longer make GET/PUT/DELETE requests.

grdsdev: Hi @jareddr thanks for pointing out the issue, a PR with the fix https://github.com/supabase/supabase-swift/pull/367

I'll get it merged and released ASAP

PR #350, opened 2024-04-19
fix: linux build

grdsdev:

- Fixes some issues building the library in Linux/Windows.
- Adds a **Supported Platforms** section to README making it clear the Linux/Windows aren't officially supported and may stop working on future versions.

See discussion in https://github.com/supabase/supabase-swift/pull/337


dshukertjr (approved):
Nice! There is one test that is failing, but everything else looks good.

Issue #77, opened 2023-04-05
Auth session doesn't refresh when using database


ryansburgoyne:
# Bug report

## Describe the bug

If my JWT has expired since I last called `supabaseClient.auth.session`, and I try to execute a command using `supabaseClient.database`, I get an error about the JWT being expired. I have to work around this by calling `_ = supabaseClient.auth.session` before every database call to force it to refresh the token.

## To Reproduce

1. Log in using Supabase Auth
2. Wait for the token to expire
3. Try to perform an operation on a database

## Expected behavior

I would expect the session to automatically be refreshed before attempting the database operation.

## System information

- OS: macOS
- Version of supabase-swift: 0.2.1

Pedr0Rocha:
At some point it was there:
https://github.com/supabase-community/supabase-swift/commit/098cc9390064ac499c1ae3a2e216cc1b3e7b8a09#diff-8ae3298b041e73e9ea309814eeb66aeb229f08265773c84cad2a9e83349ff5caR93

but then it was removed?
https://github.com/supabase-community/supabase-swift/commit/13293398a1d1fe74268c4db37f745c46977262f1#diff-8ae3298b041e73e9ea309814eeb66aeb229f08265773c84cad2a9e83349ff5caL114

kurtguenther:
I too am running into this issue (I get 401 when I restart my app after the JWT expires). I agree with @grsouza that (at least from my read of the GoTrue code) the token should already automatically refresh when `auth.session` is fetched.

https://github.com/supabase-community/gotrue-swift/blob/4bf132d73d322ee5d6d648f02deb9f2bc19cc659/Sources/GoTrue/Internal/SessionManager.swift#L41C18-L41C18

```
  func session() async throws -> Session {
    if let task = task {
      return try await task.value
    }

    guard let currentSession = try Env.localStorage.getSession() else {
      throw GoTrueError.sessionNotFound
    }

    if currentSession.isValid {
      return currentSession.session
    }

/////////// REFRESHING HERE ///////////
    task = Task {
      defer { self.task = nil }

      let session = try await Env.sessionRefresher(currentSession.session.refreshToken)
      try update(session)
      return session
    }

    return try await task!.value
  }
```

kurtguenther:
After some investigation, [this](https://github.com/supabase-community/gotrue-swift/pull/55) is what was causing the erroneous 401 for me (I had kept testing the JWT during the minute where this would have been incorrectly marked as `isValid` when it was already expired).

[...]

PR #353, opened 2024-04-23
feat(auth): auto refresh token

grdsdev:

Fix https://github.com/supabase/supabase-swift/issues/77

An access token is currently refreshed only when retrieved before calling an endpoint.

## What is the new behavior?

A part of the behavior specified above, now periodically checks for the token in the background and refreshes it upfront, the same behavior as JS and Flutter libs.


Sources/Auth/Types.swift
```diff
@@ -71,7 +71,7 @@ public struct Session: Codable, Hashable, Sendable {
 
   /// UNIX timestamp after which the ``Session/accessToken`` should be renewed by using the refresh
   /// token with the `refresh_token` grant type.
-  public var expiresAt: TimeInterval?
+  public var expiresAt: TimeInterval
```
grdsdev, line 74: @hf @kangmingtay or @J0 

Is this guarantee to always be returned in the Session? It was optional for some reason on swift.

ypotsiah:
If we have infinite lifetime for the refresh token (according to [docs](https://supabase.com/docs/guides/auth/sessions#))  seems better to verify/refresh access token just on session touch (as it was initially implemented). Nobody knows when it will be used next time so looks strange to auto refresh it by default. At least we can introduce something like:

```swift
enum SessionRefreshStrategy {
    case manual, auto
}
```

with manual option by default.

Sources/Auth/Types.swift
J0: Should be - or at least we guarantee it in the spec: https://github.com/supabase/auth/blob/master/openapi.yaml#L1920

I think we added the `ExpiresAt` after the other properties which might be why the author defined it as optional then for backward compatibility during the transition from an older Auth version to the newer one

Tests/AuthTests/Mocks/Mocks.swift
```diff
@@ -75,7 +77,7 @@ extension Session {
     accessToken: "accesstoken",
     tokenType: "bearer",
     expiresIn: 60,
-    expiresAt: Date().addingTimeInterval(60).timeIntervalSince1970,
+    expiresAt: Date().addingTimeInterval(-60).timeIntervalSince1970,
```
J0, line 80: Nice catch

J0: /nit consider adjusting `expiresIn` to -60 or 0 since we compute issuedAt from that

Sources/Auth/Internal/AutoRefreshToken.swift
```diff
@@ -0,0 +1,76 @@
...
+  @Dependency(\.logger) var logger
+
+  func start() {
+    logger?.debug("")
+
+    task?.cancel()
+    task = Task {
+      while !Task.isCancelled {
+        await autoRefreshTokenTick()
+        try? await Task.sleep(nanoseconds: UInt64(AutoRefreshToken.tickDuration) * NSEC_PER_SEC)
```
JOsacky, line 30: I was curious what your thoughts were on using this Task.sleep to precisely schedule the time that the token should be refreshed instead of checking every 30 seconds. This would match the behavior of the kotlin library.

Here is an example of an implementation: https://github.com/supabase/supabase-swift/pull/391/commits/f62567e90ef2754721728c26042b5d08024be1bc

```
 func scheduleSessionRefresh(_ session: Session) throws {
    if autoRefreshTask != nil {
      return
    }

    autoRefreshTask = Task {
      defer { autoRefreshTask = nil }

      guard let expiresAt = session.expiresAt else {
        return
      }
      let expiryDate = Date(timeIntervalSince1970: expiresAt)

      let timeIntervalToExpiry = expiryDate.timeIntervalSinceNow

      // if negative then token is expired and will refresh right away
      let timeIntervalToRefresh = max(timeIntervalToExpiry * 0.8, 0)

      try await Task.sleep(nanoseconds: UInt64(timeIntervalToRefresh * 1_000_000_000))
      let session = try await sessionRefresher.refreshSession(session.refreshToken)
    }
  }
  ```

JOsacky (commented):
Thanks for adding this change, this is exactly what we need

Sources/Auth/Internal/AutoRefreshToken.swift
grdsdev: Hi @JOsacky the 30-second interval follows the implementation of the official JS and Flutter libraries, but in Swift, we could do it the way you suggested.

I'd like input from @hf @J0 or @kangmingtay on this.

By checking every 30 seconds, we can detect sessions logged out early, this is the only thing I can think of that would benefit from the current implementation.

grdsdev:
Closed in favor of https://github.com/supabase/supabase-swift/pull/395

Sources/Auth/Internal/AutoRefreshToken.swift
JOsacky: Just seeing this comment. Thank you for thinking of the tradeoffs between the two! It seems like either would be a good solution and I saw that you chose the scheduled approach. Thank you for all the thought you put into this, I'm really excited for the next version of the swift app, I think this will greatly reduce the number of expired sessions.

Issue #364, opened 2024-05-07
Incorrect parameters extracting on PKCE flow


ypotsiah:
# Bug report

- [✔] I confirm this is a bug with Supabase, not with my own application. 
- [✔] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Seems iOS Supabase SDK incorrectly extracting URL parameters in some cases.
Example of URL: `appscheme://login-callback?code=ecd160cb-bb4a-4db2-be0c-bcf0b3abe496#_=_`

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1.  Make sure you are already logged in Facebook in mobile Safari
2. Try to link Facebook identity to some Supabase account using ASWebAuthenticationSession (prefersEphemeralWebBrowserSession = false)
4. See error of updating auth session with callback URL (invalidPKCEFlowURL)

## Expected behavior

Such URLs should be handled properly

## System information

- OS: iOS 17.4.1
- Browser: ASWebAuthenticationSession
- Version of supabase-swift: 2.8.2

## Additional context

```swift
func extractParams(from url: URL) -> [Params] {
  guard let components = URLComponents(url: url, resolvingAgainstBaseURL: false) else {
    return []
  }

  if let fragment = components.fragment {
    return extractParams(from: fragment)
  }

  if let queryItems = components.queryItems {
    return queryItems.map {
      Params(name: $0.name, value: $0.value ?? "")
    }
  }

  return []
}
```

According to the existing implementation fragment parameters will be returned immediately without merging with query items. Probably it's wrong (or maybe I just don't know the reason of such behavior)
Also not sure if it's correct from Facebook to respond with such callback URL but anyway seems we have all what we need there to proceed correctly.

grdsdev:
Thanks @ypotsiah you're right we need to extract both fragment and query, I'll PR a fix for it ASAP.

PR #365, opened 2024-05-07
fix(auth): extract both query and fragment from URL

grdsdev:

Fix https://github.com/supabase/supabase-swift/issues/364

Params extraction logic was extracting only the fragment OR the query

## What is the new behavior?

Params extraction logic now extracts both from the fragment and query if both exist.

### DIFF
--- a/Tests/AuthTests/ExtractParamsTests.swift
+++ b/Tests/AuthTests/ExtractParamsTests.swift
@@ -13,13 +13,20 @@ final class ExtractParamsTests: XCTestCase {
     let code = UUID().uuidString
     let url = URL(string: "io.supabase.flutterquickstart://login-callback/?code=\(code)")!
     let params = extractParams(from: url)
-    XCTAssertEqual(params, [Params(name: "code", value: code)])
+    XCTAssertEqual(params, ["code": code])
   }
 
   func testExtractParamsInFragment() {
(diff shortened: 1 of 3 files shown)


Tests/AuthTests/ExtractParamsTests.swift
```diff
@@ -13,13 +13,20 @@ final class ExtractParamsTests: XCTestCase {
...
 
   func testExtractParamsInFragment() {
     let code = UUID().uuidString
     let url = URL(string: "io.supabase.flutterquickstart://login-callback/#code=\(code)")!
     let params = extractParams(from: url)
-    XCTAssertEqual(params, [Params(name: "code", value: code)])
+    XCTAssertEqual(params, ["code": code])
+  }
+
+  func testExtractParamsInBothFragmentAndQuery() {
```
J0, line 26: Quick check on my understanding here: if we have both a query param and a hash param with `code` does the query param take predence?  Likewise for `error` 

(e.g.  currently does "io.supabase.flutterquickstart://login-callback/?code=123#code=abc" result in 
["code": 123 ]) 

I think with the JS lib we had a few issues related to terms which appeared in both the query and hash fragments so might be worth  confirming  against the behaviour there before merging

J0 (approved):
LGTM, great work!

PR #373, opened 2024-05-09
Export ability to get session without validation

leoMehlig:
This PR gives the client the ability to get the session without validating it first. This can be useful to display user info while offline.
The changes in `SessionManager` are required, since before the `shouldValidateExpiration` parameter was ignored if a task is already running.

### DIFF
--- a/Sources/Auth/AuthClient.swift
+++ b/Sources/Auth/AuthClient.swift
@@ -886,12 +886,25 @@ public final class AuthClient: Sendable {
     )
   }
 
+  /// Returns the current session, if any.
+  ///
+  /// The session returned by this property may be expired. Use ``session`` for a session that is guaranteed to be valid.
+  public var currentSession: Session? {
+    try? configuration.localStorage.getSession()?.session
+  }
+
+  /// Returns the current user, if any.
+  ///
+  /// The user returned by this property may be outdated. Use ``user(jwt:)`` method to get an up-to-date user instance.
+  public var currentUser: User? {
+    try? configuration.localStorage.getSession()?.session.user
+  }
+
   /// Gets the current user details if there is an existing session.
   /// - Parameter jwt: Takes in an optional access token jwt. If no jwt is provided, user() will
   /// attempt to get the jwt from the current session.
   ///
-  /// Should be used only when you require the most current user data. For faster results,
-  /// session.user is recommended.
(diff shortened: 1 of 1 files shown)


grdsdev:
Hey, I was wondering if we could maybe use something like what I'm patching below.

Flutter lib already provides properties such as `currentSession` and `currentUser`, so let's use the same API.

```patch
From a9b6b59a38b292de8ca9f4060dace7ce636a0268 Mon Sep 17 00:00:00 2001
From: Guilherme Souza <grsouza@pm.me>
Date: Thu, 9 May 2024 08:43:16 -0300
Subject: [PATCH] feat(auth): get current session without validation

---
 Sources/Auth/AuthClient.swift | 17 +++++++++++++++--
 1 file changed, 15 insertions(+), 2 deletions(-)

diff --git a/Sources/Auth/AuthClient.swift b/Sources/Auth/AuthClient.swift
index f7f67efb..ffab37e6 100644
--- a/Sources/Auth/AuthClient.swift
+++ b/Sources/Auth/AuthClient.swift
@@ -26,6 +26,20 @@ public final class AuthClient: Sendable {
       try await sessionManager.session()
     }
   }
+  
+  /// Returns the current session, if any.
+  ///
+  /// The session returned by this property may be expired. Use ``session`` for a session that is guaranteed to be valid.
+  public var currentSession: Session? {
+    try? configuration.localStorage.getSession()?.session
+  }
+
+  /// Returns the current user, if any.
+  ///
+  /// The user returned by this property may be outdated. Use ``user(jwt:)`` method to get an up-to-date user instance.
+  public var currentUser: User? {
+    try? configuration.localStorage.getSession()?.session.user
+  }
 
   /// Namespace for accessing multi-factor authentication API.
   public let mfa = AuthMFA()
@@ -890,8 +904,7 @@ public final class AuthClient: Sendable {
   /// - Parameter jwt: Takes in an optional access token jwt. If no jwt is provided, user() will
   /// attempt to get the jwt from the current session.
   ///
-  /// Should be used only when you require the most current user data. For faster results,
-  /// session.user is recommended.
+  /// Should be used only when you require the most current user data. For faster results, ``currentUser`` is recommended.
   public func user(jwt: String? = nil) async throws -> User {
     var request = HTTPRequest(url: configuration.url.appendingPathComponent("user"), method: .get)
```

leoMehlig:
@grdsdev Yes that would also work!

grdsdev (approved):

PR #375, opened 2024-05-09
fix(auth): sign out regardless of request success

leoMehlig:
This lets the client reset the session, which is relevant if a logout fails (for example if the user is offline).

### DIFF
--- a/Sources/Auth/AuthClient.swift
+++ b/Sources/Auth/AuthClient.swift
@@ -698,15 +698,14 @@ public final class AuthClient: Sendable {
   /// If using ``SignOutScope/others`` scope, no ``AuthChangeEvent/signedOut`` event is fired.
   /// - Parameter scope: Specifies which sessions should be logged out.
   public func signOut(scope: SignOutScope = .global) async throws {
-    let accessToken = currentSession?.accessToken
+    guard let accessToken = currentSession?.accessToken else {
+        configuration.logger?.warning("signOut called without a session")
+        return
+    }
 
     if scope != .others {
         await sessionManager.remove()
         eventEmitter.emit(.signedOut, session: nil)
-    }
-
-    guard let accessToken else {
-        throw AuthError.sessionNotFound
     }
 
     do {
--- a/Sources/Auth/AuthClient.swift
+++ b/Sources/Auth/AuthClient.swift
@@ -698,15 +698,24 @@ public final class AuthClient: Sendable {
   /// If using ``SignOutScope/others`` scope, no ``AuthChangeEvent/signedOut`` event is fired.
   /// - Parameter scope: Specifies which sessions should be logged out.
   public func signOut(scope: SignOutScope = .global) async throws {
-    do {
-      // Make sure we have a valid session.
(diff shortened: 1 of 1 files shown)


grdsdev:
@leoMehlig if I understand this is used to fix an issue with the `signOut` not working when offline.

Looking at the flutter implementation again https://github.com/supabase/supabase-flutter/blob/main/packages/gotrue/lib/src/gotrue_client.dart#L844 it resets the session before making the request, so it will sign out regardless of the request succeeding or not. I think this is the way to go here.

Now that we have the `currentSession` without validating the session validity, it should be easy to implement the same behavior as the Flutter lib.

Let me know if you want to do it in this PR, or I should PR this changes I mentioned.

leoMehlig:
Good point @grdsdev! I've adjusted the PR to do this and removed the `resetSession` function. 

I had to use the `execute` instead of the `authorizedExecute` function and manually add the auth header. I think this is fine, but we could also refactor the `authorizedExecute` to allow passing in an accessToken.

Sources/Auth/AuthClient.swift
```diff
@@ -698,15 +698,24 @@ public final class AuthClient: Sendable {
...
+    let accessToken = currentSession?.accessToken
+
+    if scope != .others {
+        await sessionManager.remove()
+        eventEmitter.emit(.signedOut, session: nil)
+    }
+
+    guard let accessToken else {
+        throw AuthError.sessionNotFound
+    }
```
grdsdev, line 710: Maybe we should not throw in case there isn't a session, and just log a warning and return.
That will also make sure we don't emit the signedOut event even when there is no session.

```suggestion
    guard let accessToken = currentSession?.accessToken else {
        configuration.logger?.warning("signOut called without a session")
        return
    }

    if scope != .others {
        await sessionManager.remove()
        eventEmitter.emit(.signedOut, session: nil)
    }
```

grdsdev (changes requested):

Sources/Auth/AuthClient.swift
grdsdev: @leoMehlig this suggestion is different than the flutter implementation.

Tagging @dshukertjr to discuss which behavior we should go with, 

1. Flutter sign out emits the signed out event even if there's no current session.
2. I'm suggesting we don't emit the event in that case

leoMehlig: I'm also fine to go with the suggestions. The only case this could go against expected behavior is when trying to log out all (or other sessions).

dshukertjr: Yeah, I like this suggestion of not emitting `signOut` event when there is no access token!

grdsdev: @leoMehlig let's go with https://github.com/supabase/supabase-swift/pull/375#discussion_r1595489030 then.

leoMehlig: Perfect! Updated it.

grdsdev (approved):

PR #377, opened 2024-05-10
feat: expose Realtime options on SupabaseClient

bryandubno:

Exposes public Realtime configuration options.

Realtime configuration options are public, but cannot be configured by the client.

## What is the new behavior?

Realtime configuration options can now be set by the client.

## Additional context

This is not ready to be merged. Tests need to be updated.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -158,11 +158,18 @@ let package = Package(
         "Functions",
       ]
     ),
-    .testTarget(name: "SupabaseTests", dependencies: ["Supabase"]),
+    .testTarget(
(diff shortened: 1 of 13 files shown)


grdsdev:
@bryandubno 
> I do also wonder if it would be better to add to SupabaseClientOptions instead?

Yes please, let's add it to SupabaseClientOptions under a `realtime` attribute.

grdsdev:
@bryandubno I've made a few changes to your code and also updated tests.

grdsdev (approved):

PR #379, opened 2024-05-12
fix(auth): header being overridden

leoMehlig:
I noticed that auth calls like `updateUser` would fail due to a wrong JWT token. The cause for this was, that the `execute` method would override the `Authentication` header to the anon key. I changed the call, such that the `execute` method does not override existing headers.

### DIFF
--- a/Sources/Auth/Internal/APIClient.swift
+++ b/Sources/Auth/Internal/APIClient.swift
@@ -23,7 +23,7 @@ struct APIClient: Sendable {
 
   func execute(_ request: HTTPRequest) async throws -> HTTPResponse {
     var request = request
-    request.headers.merge(with: HTTPHeaders(configuration.headers))
+    request.headers = HTTPHeaders(configuration.headers).merged(with: request.headers)
 
     let response = try await http.send(request)
 


grdsdev:
Hi @leoMehlig I don't see this issue happening on my tests, do you have a reproducible example?

leoMehlig:
Does it not override the `session.accessToken` with the API key for you?

grdsdev:
> Does it not override the session.accessToken with the API key for you?

It doesn't, I'm verifying with the following test: https://github.com/supabase/supabase-swift/blob/main/Tests/IntegrationTests/AuthClientIntegrationTests.swift#L139

leoMehlig:
> > Does it not override the session.accessToken with the API key for you?
> 
> It doesn't, I'm verifying with the following test: https://github.com/supabase/supabase-swift/blob/main/Tests/IntegrationTests/AuthClientIntegrationTests.swift#L139

I think the difference in the test is, that it does not have a `Authorization` header set in the text, but in production this is set to the apiKey.

grdsdev (approved):

Issue #385, opened 2024-05-15
App crashes with error: pointer being freed was not allocated


kevlust:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

My app will occasionally crash on startup due to error: "pointer being freed was not allocated". It only happens sometimes, which makes it harder to reproduce. To be completely honest, I'm not sure if this is a bug on the Supabase package or on my own code. Happens when I try to access `storage` from within the client object. Attaching screenshots of the error.

## Screenshots

<img width="1904" alt="Screenshot 2024-05-14 at 9 37 12 PM" src="https://github.com/supabase/supabase-swift/assets/16130851/8a563904-fae8-4ec9-b60d-fd3a3a5ad22c">
<img width="1904" alt="Screenshot 2024-05-14 at 9 36 43 PM" src="https://github.com/supabase/supabase-swift/assets/16130851/de3489fd-a501-4f8d-a3c0-5fe8e65d56f0">

## System information

- iOS 17.4.1
- Supabase v2.9.0

## Additional context

Here's some screenshots from an older crash I saw before updating to 2.9.0. It seems to be related:
<img width="1904" alt="Screenshot 2024-05-10 at 7 38 01 PM" src="https://github.com/supabase/supabase-swift/assets/16130851/56d6c970-fc14-4168-997f-9e7654fbee2b">
<img width="1904" alt="Screenshot 2024-05-10 at 7 37 50 PM" src="https://github.com/supabase/supabase-swift/assets/16130851/66aed2e3-30c6-44b8-acf6-571bfe520301">

Issue #239, opened 2024-02-10
Random crash from Supabase code right after app launch


kevlust:
# Bug report

## Describe the bug

My iOS app that uses this package crashed a few seconds after launch. Upon reading through the crash stack trace, I noticed that this is happening due to some code from this package.

This happens very randomly (I can't manually reproduce this crash). However, I think (**not 100% sure**) this has happened when testing the app with bad reception, so that might be helpful here. I'll keep trying to reproduce this.

## System information

- iOS: 17.2.1
- iPhone 14 Pro
- Supabase 2.2.1
- KeychainAccess 4.2.2

## Additional context

```
Crashed: com.apple.root.user-initiated-qos.cooperative
0  libsystem_kernel.dylib         0x9fbc __pthread_kill + 8
1  libsystem_pthread.dylib        0x5680 pthread_kill + 268
2  libsystem_c.dylib              0x75b90 abort + 180
3  libsystem_malloc.dylib         0x26200 malloc_vreport + 908
4  libsystem_malloc.dylib         0x264a8 malloc_zone_error + 104
5  libsystem_malloc.dylib         0x1dac8 nanov2_guard_corruption_detected + 44
6  libsystem_malloc.dylib         0x7e78 nanov2_find_block_and_allocate + 402
7  libswiftCore.dylib             0x3a2c80 swift::swift_slowAllocTyped(unsigned long, unsigned long, unsigned long long) + 104
8  libswiftCore.dylib             0x3a7888 _swift_allocObject_ + 1088
9  libswiftCore.dylib             0x33bfcc specialized static _DictionaryStorage.resize(original:capacity:move:) + 372
10 libswiftCore.dylib             0x17ae08 _NativeDictionary._copyOrMoveAndResize(capacity:moveElements:) + 300
11 libswiftCore.dylib             0xd1e34 _NativeDictionary.setValue(_:forKey:isUnique:) + 272
12 libswiftCore.dylib             0xab6d8 Dictionary._Variant.setValue(_:forKey:) + 228
13 libswiftCore.dylib             0x58624 Dictionary.subscript.setter + 512
14 MyApp                          0xf9cf78 Options.query(ignoringAttributeSynchronizable:) + 1331 (Keychain.swift:1331)
15 MyApp                          0xf9c694 Keychain.getData(_:ignoringAttributeSynchronizable:) + 611 (Keychain.swift:611)
[...]

PR #386, opened 2024-05-15
fix: race condition when accessing SupabaseClient

grdsdev:

I wasn't able to reproduce it, but I have an assumption that these changes may fix the following crashes.

Fix https://github.com/supabase/supabase-swift/issues/239
Fix https://github.com/supabase/supabase-swift/issues/385

Inner clients from SupabaseClient were defined as `lazy var`, but `lazy var` isn't thread-safe in Swift, this was probably causing some crashes when accessed from different threads.

We also relied on it to mark the `SupabaseClient` class as `@unchecked Sendable`.

## What is the new behavior?

Moved inner clients to a `LockIsolated` and made `SupabaseClient` conform to `Sendable`.

### DIFF
--- a/Sources/Supabase/Deprecated.swift
+++ b/Sources/Supabase/Deprecated.swift
@@ -1,0 +1,26 @@
+//
+//  Deprecated.swift
+//
+//
+//  Created by Guilherme Souza on 15/05/24.
+//
+
+import Foundation
+
+extension SupabaseClient {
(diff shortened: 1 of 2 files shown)


dshukertjr (approved):

kevlust:
@grdsdev you're awesome. Thanks for coming up with this fix so quickly!

PR #391, opened 2024-05-16
Schedule session refresh at 20% of time before expiry

JOsacky:

This change makes it so that we schedule a session refresh refresh at 20% of time before expiry. This matches the kotlin library behavior:
https://github.com/supabase-community/supabase-kt/blob/b39b2fcd0320be7d26a481d3dc93d3e955b37000/GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/AuthImpl.kt#L423-L430

This PR fixes https://github.com/supabase/supabase-swift/issues/77

Current behavior doesn't have an auto refresh mechanism, it is lazily refreshed as far as I can tell if the session is expired when calling `SessionManager.session`. Currently in apps that use the swift supabase library to get the JWT token and use for auth for non supabase APIs, `SessionManager.session` may not be called. There also seems to be other cases in the issue referenced above where refresh isn't called in time. 

## What is the new behavior?

We will schedule a task to refresh the token at 20% of time before expiry to match the kotlin library's functionality

https://github.com/supabase-community/supabase-kt/blob/b39b2fcd0320be7d26a481d3dc93d3e955b37000/GoTrue/src/commonMain/kotlin/io/github/jan/supabase/gotrue/AuthImpl.kt#L394-L401

## Additional context

Tested with my personal app and verified that the refresh is scheduled at the right time and refreshes the token properly

### DIFF
--- a/Sources/Auth/Internal/SessionManager.swift
+++ b/Sources/Auth/Internal/SessionManager.swift
@@ -7,6 +7,7 @@ struct SessionRefresher: Sendable {
 
 actor SessionManager {
   private var task: Task<Session, any Error>?
+  private var autoRefreshTask: Task<Void, any Error>?
 
   private var storage: any AuthLocalStorage {
     Current.configuration.localStorage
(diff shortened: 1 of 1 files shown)


JOsacky:
I see there is already a PR for this closing !

https://github.com/supabase/supabase-swift/pull/353/files

PR #420, opened 2024-06-13
Update RealtimeChannelV2.swift

bryandubno:

Adds missing `.unsubscribed` status change

Once `.unsubscribing` from a channel occurs, `.unsubscribed` never gets set (leaving `.unsubscribing` indefinitely as the status)

## What is the new behavior?

Status is changed to `.unsubscribed`


grdsdev:
Nice catch @bryandubno , thanks for the fix.

grdsdev (approved):

PR #428, opened 2024-06-27
fix: concurrency warnings pre swift 6 support

grdsdev:

This is part of the Swift 6 mode support, some concurrency warnings will become errors when Swift 6 is enabled.

This PR fixes a couple of concurrency warnings, by not using the `ISO8601DateFormatter`, which isn't `Sendable` in favor of the `DateFormatter`, which is `Sendable` and is already being used on some places

### DIFF
--- a/Sources/Auth/Defaults.swift
+++ b/Sources/Auth/Defaults.swift
@@ -9,25 +9,13 @@ import Foundation
 import Helpers
 
 extension AuthClient.Configuration {
-  private static let dateFormatterWithFractionalSeconds = { () -> ISO8601DateFormatter in
-    let formatter = ISO8601DateFormatter()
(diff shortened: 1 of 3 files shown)


dshukertjr (approved):

dpelletier2017:
I tried the new version, but this change seems to have broken the API completely.
SELECT and INSERT don't work anymore. For example, I get this error :

Swift.DecodingError.dataCorrupted(Swift.DecodingError.Context(codingPath: [_JSONKey(stringValue: "Index 0", intValue: 0), CodingKeys(stringValue: "created_at", intValue: nil)], debugDescription: "Invalid date format: 2024-06-15T18:12:04+00:00", underlyingError: nil))

Although the date string in the error output seems to be OK.

I'm using Xcode 15.4, with Swift 5.

PR #430, opened 2024-06-27
fix(auth): use project ref as namespace for storing token

grdsdev:

Bug fix

When using multiple projects, for example, a dev and a production, both environments shares the same token.

## What is the new behavior?

- Use project ref as a namespace when accessing stored token.
- Add support for providing a custom storage key to store the token on local storage.

### DIFF
--- a/Sources/Auth/AuthClientConfiguration.swift
+++ b/Sources/Auth/AuthClientConfiguration.swift
@@ -24,6 +24,9 @@ extension AuthClient {
     public var headers: [String: String]
     public let flowType: AuthFlowType
     public let redirectToURL: URL?
+
+    /// Optional key name used for storing tokens in local storage.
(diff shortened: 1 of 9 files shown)


dshukertjr (commented):
If a user using a version of supabase-swift prior to to this update updates to a new version with this update included, would they be signed out?

grdsdev:
> If a user using a version of supabase-swift prior to to this update updates to a new version with this update included, would they be signed out?

No, it migrates the token to the new key: https://github.com/supabase/supabase-swift/pull/430/files#diff-6132ac0f2c100e8db5a2725259009c843e6dff71e699aadfcdf61f4b01c00192R30-R40

dshukertjr (approved):
Sorry, missed the migration code! LGTM!

PR #442, opened 2024-07-04
fix: Add private topic to Realtime

filipecabaco:

Add private topic to Realtime

### DIFF
--- a/Sources/Realtime/V2/RealtimeChannelV2.swift
+++ b/Sources/Realtime/V2/RealtimeChannelV2.swift
@@ -12,7 +12,7 @@ import Helpers
 public struct RealtimeChannelConfig: Sendable {
   public var broadcast: BroadcastJoinConfig
   public var presence: PresenceJoinConfig
-  public var privateTopic: Bool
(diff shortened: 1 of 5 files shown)


Sources/Realtime/V2/RealtimeChannelV2.swift
```diff
@@ -12,6 +12,7 @@ import Helpers
 public struct RealtimeChannelConfig: Sendable {
   public var broadcast: BroadcastJoinConfig
   public var presence: PresenceJoinConfig
+  public var privateTopic: Bool
```
grdsdev, line 15: Although `private` is a swift keyword, you can use it as a property name if you escape it with `, example:

```suggestion
  public var `private`: Bool
```

Or maybe we could name it `isPrivate` to avoid the escaping?

filipecabaco: I like that! isPrivate is a great name that I should have thought about 😅

grdsdev (approved):

Issue #441, opened 2024-07-02
[Auth] Need API to allow for getting a Session without actually signing in/persisting via the client


MilesV64:
My use case is an app where a user with a phone account can migrate to an apple sign-in account, and we do so by first getting a valid Apple session, then calling a migrate function on our server using both the current user session and the new Apple session, and then if that succeeds, logging into the apple account with `auth.setSession` using the apple session.

We can't work around this by creating an independent AuthClient like so:

```
// create a new AuthClient, so in case migration fails we don't log out the user
let independentAuthClient = AuthClient(
    url: supabaseURL.appendingPathComponent("/auth/v1"),
    localStorage: InMemoryAuthStorage(), // Subclass which does not persist the session
    logger: nil
)
newUserSession = try await independentAuthClient.signInWithIdToken(credentials: .init(
    provider: .apple,
    idToken: token
))
```

Because it turns out AuthClient init actually points to a singleton.

Previously I forked the library to include a method `getSessionFromIdToken` which does the same API call as `signInWithIdToken` but then simply returns the `Session` rather than persisting it and "logging in" client side, but we'd like to see that officially supported.

Alternatively, support for multiple AuthClients (not pointing to a single dependency) would help too.

P.S. thank you for the quick turnaround on my last issue!

EDIT: Made a mistake with the original post thinking a function previously existed in the library.

grdsdev:
Hi @MilesV64 

Sorry for the late reply, but I was a bit confused because I couldn't remember ever having this `getSessionFromIdToken` method in the library 😅, your EDIT made things clearer.

Supporting multiple client instances is something that I want to tackle for a while, instead of adding this method, I prefer adding support for multiple instances.

I'll work on this next.

grdsdev:
Started to work on a PR for this: https://github.com/supabase/supabase-swift/pull/445

MilesV64:
Thank you @grdsdev! And yea, my mistake with the original post and appreciate you working on multiple instances! 

One nice-to-have would be being able to access the `defaultHeaders` of `supabaseClient` (currently it's internal) so we don't need to hard code those headers in the newly created auth instance.

MilesV64:
Hi, thanks a lot for this! Sorry for just now finding this, but there's an issue: when creating a new AuthClient, it still uses the shared `eventEmitter` which is causing bugs due to the new authClient sending events which I cannot differentiate. Link to comment on the PR https://github.com/supabase/supabase-swift/pull/445/files#r1714098710

PR #445, opened 2024-07-05
feat(auth): add support for multiple auth instances

grdsdev:

Close https://github.com/supabase/supabase-swift/issues/441

`Auth` uses a singleton dependency, so all client instances use the same set of dependencies, this makes it impossible to use multiple auth instances properly.

## What is the new behavior?

`AuthClient` has a unique identifier, now used to index an instance of the `Dependencies` struct. Each `AuthClient` has its own `Dependencies`.

### DIFF
--- a/Sources/Auth/AuthClient.swift
+++ b/Sources/Auth/AuthClient.swift
@@ -13,7 +13,7 @@ import Helpers
 typealias AuthClientID = UUID
 
 public final class AuthClient: Sendable {
-  private let clientID = AuthClientID()
+  let clientID = AuthClientID()
 
   private var api: APIClient { Dependencies[clientID].api }
   var configuration: AuthClient.Configuration { Dependencies[clientID].configuration }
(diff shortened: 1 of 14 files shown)


dshukertjr (approved):

Sources/Auth/AuthClient.swift
```diff
@@ -33,31 +37,39 @@ public final class AuthClient: Sendable {
...
+    AuthAdmin(clientID: clientID)
+  }
 
   /// Initializes a AuthClient with a specific configuration.
   ///
   /// - Parameters:
   ///   - configuration: The client configuration.
   public init(configuration: Configuration) {
-    Current = Dependencies(
+    Dependencies[clientID] = Dependencies(
```
MilesV64, line 67: This should also use its own eventEmitter instead of eventEmitter.shared -- currently all auth event changes are passed to the same stream without any way of differentiating.

grdsdev: You're right, thanks for bringing up the issue.

I'll PR a fix soon.

Thanks again.

PR #454, opened 2024-07-10
fix(storage): list folders

grdsdev:

Bug fix

`Bucket` and `FileObject` structs consider some optional fields as required, this breaks in cases where a _folder_ is returned from the `list` method, as folders have only the `name` as non-null.

## Additional context

Close https://github.com/supabase/supabase-swift/issues/479


Sources/Storage/FileObject.swift
```diff
@@ -2,26 +2,24 @@ import Foundation
...
-  public var name: String
+  public var name: String?
   public var bucketId: String?
   public var owner: String?
-  public var id: String
-  public var updatedAt: Date
-  public var createdAt: Date
-  public var lastAccessedAt: Date
-  public var metadata: [String: AnyJSON]
-  public var buckets: Bucket?
```
grdsdev, line 13: from my tests this is never returned

Sources/Storage/FileObject.swift
```diff
@@ -2,26 +2,24 @@ import Foundation
...
+    name: String? = nil,
     bucketId: String? = nil,
     owner: String? = nil,
-    id: String,
-    updatedAt: Date,
-    createdAt: Date,
-    lastAccessedAt: Date,
-    metadata: [String: AnyJSON],
-    buckets: Bucket? = nil
+    id: UUID,
```
Axxou, line 18: On my side, its only after I put the `id: UUID? = nil` that the fetch worked.

Sources/Storage/FileObject.swift
```diff
@@ -2,26 +2,24 @@ import Foundation
...
+  public var name: String?
   public var bucketId: String?
   public var owner: String?
-  public var id: String
-  public var updatedAt: Date
-  public var createdAt: Date
-  public var lastAccessedAt: Date
-  public var metadata: [String: AnyJSON]
-  public var buckets: Bucket?
+  public var id: UUID
```
Axxou, line 8: To complete with my comment about optional id : `public var id: UUID?`

grdsdev:
@fenos does this make sense from a backend perspective?

All but `name` field from `FileObject` struct are optional?

dshukertjr (approved):

PR #455, opened 2024-07-11
fix(auth): prevent from requesting login keychain password os macOS

ekurutepe:

This PR aligns the behavior of KeychainLocalStorage on iOS and macOS. 

When one app save an item in KeychainLocalStorage with an access group on macOS, another app/extension with the same access group would not be able to access the item without the user having to enter their login keychain password. On iOS this would work without user having to enter any passwords.

## What is the new behavior?

Both on iOS and macOS apps/extensions are able to access keychain items with the shared access group correctly without the user having to enter their password.

### DIFF
--- a/Sources/Auth/Internal/Keychain.swift
+++ b/Sources/Auth/Internal/Keychain.swift
@@ -67,6 +67,11 @@
         query[kSecAttrAccessGroup as String] = accessGroup
       }
 
+      // this is highly recommended for all keychain operations and makes the
+      // macOS keychain item behave like an iOS keychain item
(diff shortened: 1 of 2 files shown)


grdsdev (approved):

ekurutepe:
that was quick! thanks. 🎉

grdsdev:
Hey @ekurutepe getting back to you since you proposed this change, some people are having issues after this change, when using without an access group as you can follow on https://github.com/supabase/supabase-swift/issues/516

Do you have any idea why, and a fix for it?

PR #475, opened 2024-07-24
Provide a means of synchronously invalidating local session data

bryandubno:
When re-installing an app, I want previous session data that was persisting in the keychain to be deleted as well. I'm not sure this is the best approach, but it allows for a synchronous removal of any existing session data.

### DIFF
--- a/Sources/Auth/AuthClient.swift
+++ b/Sources/Auth/AuthClient.swift
@@ -778,6 +778,11 @@ public final class AuthClient: Sendable {
     return session
   }
 
+  /// Provides a synchronous means of instantly invalidating the current local session.
+  public func invalidateCurrentSession() {
+    try? sessionStorage.delete()
+  }
+
   /// Signs out the current user, if there is a logged in user.
   ///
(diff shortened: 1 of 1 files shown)


grdsdev:
Hi @bryandubno

I'm reluctant to add new methods to the library that don't exist on JS and Flutter, before I evaluate this added method, let's make sure you can't use any other approach.

The `signOut` method invalidates the session before sending the backend request, I know it isn't sync but it may work depending on your scenario.

```swift
Task {
  try? await supabase.auth.signOut()
}
```

Another approach is to implement your own `AuthLocalStorage` with the logic to remove Keychain data when a re-install is detected.

```swift
struct MyCustomAuthStorage: AuthLocalStorage {

  init(keychain: Keychain, userDefaults: UserDefaults) {
    self.keychain = keychain
    if userDefaults.bool(forKey: "hasRunBefore") == false {
      userDefaults.set(true, forKey: "hasRunBefore")
      keychain.removeAllData()
    }
  }

  // AuthLocalStorage methods implementation...
}
```

Let me know if you can try any of those approaches.

PR #487, opened 2024-08-02
Corrected project name.

chintan100:

Minor typo.

It mentions project name as Firebase.

## What is the new behavior?

Corrected project name to Supabase.

## Additional context

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,6 +88,6 @@ let client = SupabaseClient(
 
 ## Sponsors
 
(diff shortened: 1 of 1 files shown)


mergesort:
I think the text was previously correct, since they're saying that they're building the features that Firebase has to offer, but as an open source alternative.

grdsdev:
What @mergesort said is correct, we're stating that we're building Firebase's features using open source technologies.

PR #493, opened 2024-08-06
fix(auth): add missing figma, kakao, linkedin_oidc, slack_oidc, zoom, and fly providers

grdsdev:

Fix adding missing auth providers


J0 (approved):
Thank you!

PR #502, opened 2024-08-13
fix(auth): store code verifier in keychain

grdsdev:

Bug fix

Auth client stores code verifier for PKCE flow in memory only

## What is the new behavior?

Stores code verifier in keychain to avoid any issues with lost of code in PKCE flow


J0 (approved):
Thanks!

PR #514, opened 2024-08-26
feat(postgrest): use `Date` when filtering columns

LucasAbijmil:
When using filters on `Timestampz` typed columns, the `Date.now.description` declaration only works for users with time in 24-hour format and not for those using 12-hour one. 
This pull request makes using filters simpler and less error-prone when manipulating the `Date` type.

In order to use filters with `Date` type, we need to have a custom logic in order to make it works properly.

## What is the new behavior?
When using a filter, the `Date.now` (or `Date()`) works by the book without having to implement a custom logic.

## Additional context
If my change doesn't work on other Date/Time types, please let me know. I would be happy to make this works for everyone and every Date/Time type

### DIFF
--- a/Sources/PostgREST/URLQueryRepresentable.swift
+++ b/Sources/PostgREST/URLQueryRepresentable.swift
@@ -27,6 +27,14 @@ extension UUID: URLQueryRepresentable {
   public var queryValue: String { uuidString }
 }
 
+extension Date: URLQueryRepresentable {
+  public var queryValue: String {
+    let formatter = ISO8601DateFormatter()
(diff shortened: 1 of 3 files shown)


grdsdev:
That is a very good improvement, let me make a few tests using other time/date types to see how it behaves.

LucasAbijmil:
> That is a very good improvement, let me make a few tests using other time/date types to see how it behaves.

Thank you very much! I hope we can find a solution to make things easier for everyone :)

grdsdev (approved):

PR #518, opened 2024-08-29
feat(auth): add support for error codes and refactor `AuthError`

grdsdev:

This PR includes some changes on Auth error system, I tried my best to keep it compatible and without breaking changes.

If you see anything that could be a breaking change, please let me know.

### Error codes

Introduce `ErrorCode` type as a raw representable string, added static properties for each of the known error codes. Since error codes are raw representable strings, if an unknown error code is returned, decoding is going to work fine.

### Deprecated errors

- `missingExpClaim` error deprecated, it is never thrown now, but I kept it for retro-compatibility 
- `malformedJWT` error deprecated, it is never thrown now, but I kept  it for retro-compatibility 
- `sessionNotFound` error deprecated, use `sessionMissing` instead
- `pkce(_:)` error deprecated, use `pkceGrantCodeExchange` instead
- `invalidImplicitGrantFlowURL` error deprecated, use `implicitGrantRedirect` instead
- `missingURL` error deprecated, it is never thrown, but I kept it for retro-compatibility
- `api(_:)` error deprecated, use `api(message:errorCode:underlyingData:underlyingResponse:)` instead

### Added errors

- `sessionMissing` thrown when a session is required to proceed, but none was found, either thrown by the client, or returned by the server.
- `weakPassword` thrown when password is deemed weak, check associated reasons to know why.
- `api(message:errorCode:underlyingData:underlyingResponse:)` thrown by API when an error occurs, check `errorCode` to know more, or use `underlyingData` or `underlyingResponse` for access to the response which originated this error.
- `pkceGrantCodeExchange` thrown when an error happens during PKCE grant flow.
- `implicitGrantRedirect` thrown when an error happens during implicit grant flow.

### DIFF
--- a/Sources/Auth/Internal/SessionManager.swift
+++ b/Sources/Auth/Internal/SessionManager.swift
@@ -40,7 +40,7 @@ private actor LiveSessionManager {
   func session() async throws -> Session {
     try await trace(using: logger) {
       guard let currentSession = try sessionStorage.get() else {
-        throw SupabaseAuthSessionMissingError()
+        throw AuthError.sessionMissing
       }
 
       if !currentSession.isExpired {
--- a/Sources/Auth/Internal/SessionManager.swift
+++ b/Sources/Auth/Internal/SessionManager.swift
@@ -40,7 +40,7 @@ private actor LiveSessionManager {
   func session() async throws -> Session {
     try await trace(using: logger) {
       guard let currentSession = try sessionStorage.get() else {
-        throw AuthError.sessionNotFound
+        throw SupabaseAuthSessionMissingError()
       }
 
       if !currentSession.isExpired {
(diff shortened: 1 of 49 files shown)


Sources/Auth/Internal/SessionManager.swift
```diff
@@ -40,7 +40,7 @@ private actor LiveSessionManager {
   func session() async throws -> Session {
     try await trace(using: logger) {
       guard let currentSession = try sessionStorage.get() else {
-        throw AuthError.sessionNotFound
+        throw AuthError.sessionMissing
```
dshukertjr, line 43: I feel like this could technically be considered a breaking change, but I will leave it up to you what to do with it.

dshukertjr (approved):

Sources/Auth/Internal/SessionManager.swift
grdsdev: Not really since I added a `static let sessionNotFound: AuthError { .sessionMissing }`, and deprecated it.

In a Swift enum, `cases` and `static let` are "kinda" equivalent.

Issue #486, opened 2024-08-01
"Refresh Token Not Found" = automatic logout after 2-3 days after login


EduardMe:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

About 2-3 days after logging in, the user gets automatically logged out (we are using the OTP login and use the API on iOS and macOS). Checking the error messages, we see:

`api(Auth.AuthError.APIError(msg: nil, code: nil, error: Optional("invalid_grant"), errorDescription: Optional("Invalid Refresh Token: Refresh Token Not Found"), weakPassword: nil))`

## To Reproduce

Using the Swift API of Supabase, login, use the app for 2-3 days and you will get logged out.

## Expected behavior

User should stay logged in indefinitely or as per the settings (see additional information).

## System information

- OS: macOS, iOS
- Version of supabase-swift: 2.15.2 

## Additional context

We have tried to increase the rate limits:
Token Refreshes = 184
Token Verifications = 100
Sign ups and sign ins = 50
Time-box user sessions = 0 
Inactivity timeout = 0
Access Tokens expiry time = 10800 (3 hours)
Refresh token reuse interval = 30

Let me know if the settings can cause this.

hf:
This is likely due to a bug in your application that refreshes the token in parallel. Please see this doc for more information about such issues: https://supabase.com/docs/guides/auth/sessions#what-is-refresh-token-reuse-detection-and-what-does-it-protect-from

hf:
Also be aware that if the user signed out from another device or platform (web) they are signed out from everywhere. You can control this with a parameter.

EduardMe:
Ok, after about 2-3 days of testing with the changes, I'm still getting logged out automatically on macOS (2x in the last 2-3 days) with the error message:

```
api(Auth.AuthError.APIError(msg: nil, code: nil, error: Optional("invalid_grant"), errorDescription: Optional("Invalid Refresh Token: Already Used"), weakPassword: nil))
```

I haven't hit the logout button anywhere during testing. 

Any ideas how to fix it or if it's a bug in the library?

grdsdev:
Hi @EduardMe 

do you have any extension also using Supabase? Widgets...

EduardMe:
[...]

PR #561, opened 2024-10-10
fix: general auth improvements

grdsdev:

Several auth improvements, more details below.

fix(auth): remove session in case token refresh fails and fire `SIGNED_OUT` event
fix(auth): deprecate `emailChangeToken` field from `UserAttributes` type

fix(auth): auto refresh token in background
Add both `startAutoRefresh` and `stopAutoRefresh` methods in `AuthClient`, caller doesn't need to manually call those methods if client is configured with `autoRefreshToken: true` and platform supports observing app lifecycle changes (i.e. all Apple platforms). Use those methods if you want have full control over when auto refresh background process runs.

fix: trigger listeners in order
The order which listeners were called were non-deterministic, as it was backed by a `Dictionary`, I changed it to an Array to it keeps the order in which they are attached.

This means that if you have:
```swift
client.auth.onAuthStateChange { _, _ in 
  print("1")
}

client.auth.onAuthStateChange { _, _ in 
  print("2")
}
```
And a new event is triggered, you now always get:
```
1
2
```
Printed to console in order.

### Additional context
Tries to workaround the https://github.com/supabase/supabase-swift/issues/486 issue

### DIFF
--- a/Sources/Auth/Internal/SessionManager.swift
+++ b/Sources/Auth/Internal/SessionManager.swift
@@ -84,7 +84,7 @@ private actor LiveSessionManager {
                 url: configuration.url.appendingPathComponent("token"),
                 method: .post,
                 query: [
-                  URLQueryItem(name: "grant_type", value: "refresh_token"),
+                  URLQueryItem(name: "grant_type", value: "refresh_token")
                 ],
                 body: configuration.encoder.encode(
                   UserCredentials(refreshToken: refreshToken)
@@ -97,10 +97,13 @@ private actor LiveSessionManager {
             eventEmitter.emit(.tokenRefreshed, session: session)
 
             return session
+          } catch let error as RetryableError {
+            logger?.debug("Refresh token failed with a retryable error: \(error)")
+            throw error
           } catch {
-            logger?.debug("Failed to refresh token: \(error)")
+            logger?.debug("Refresh token failed with a non-retryable error: \(error)")
+
             remove()
-            eventEmitter.emit(.signedOut, session: nil)
 
             throw error
           }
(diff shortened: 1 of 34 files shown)


Sources/Auth/Internal/SessionManager.swift
```diff
@@ -55,40 +55,50 @@ private actor LiveSessionManager {
...
+            await scheduleNextTokenRefresh(session)
 
-          return session
+            return session
+          } catch {
+            logger?.debug("Failed to refresh token: \(error)")
+            remove()
+            eventEmitter.emit(.signedOut, session: nil)
+
+            throw error
```
grdsdev, line 100: The HTTP client already handles request retries, so by this point token refresh was already retried a couple of times, we can remove it right away.

Sources/Auth/Internal/SessionManager.swift
```diff
@@ -39,56 +43,67 @@ private actor LiveSessionManager {
...
-          update(session)
-          eventEmitter.emit(.tokenRefreshed, session: session)
+            update(session)
+            eventEmitter.emit(.tokenRefreshed, session: session)
 
-          await scheduleNextTokenRefresh(session)
+            return session
+          } catch {
+            logger?.debug("Failed to refresh token: \(error)")
+            remove()
```
dshukertjr, line 102: Do we want to remove the token even if it's a network failure? Should we only remove it if it's a certain error code(s)?

grdsdev: auth-js removes it if it isn't a retryable error https://github.com/supabase/auth-js/blob/master/src/GoTrueClient.ts#L1974-L1977

since the HTTP client retries the request a few times before reaching this point, the behavior is roughly the same as auth-js

How does flutter works?

hf: No, network failure means there's no internet. Users shouldn't be logged out ever from this.

hf: Ideally supplement the guides (and we should do this) to have the device detect there's likely no internet (Airplane mode, no Wifi, no Data / Cell service) and stop auto refreshes while this is true, as this would likely lead in a network error.

In the web it's a bit hard to detect this but I'll add it to auth-js.

grdsdev: @hf in which cases should user be logged out? if any?

grdsdev: I changed it so session is deleted only if error is non-retryable, just like JS and Flutter.

dshukertjr (approved):

PR #564, opened 2024-10-14
Replace to HTTPTypes Components from Helpers Components

zunda-pixel:
### HTTP Header

- Remove Helpers.HTTPHeader
- Add HTTPTypes.HTTPFields

### HTTP Method

- Remove Helpers.HTTPMethod
- Add HTTPTypes.HTTPRequest.Method


grdsdev:
Hi @zunda-pixel thats a nice change, I was waiting for adopting `HTTPTypes` until they found a way of representing the `body` too, then I'd also adopt `HTTPRequest` from `HTTPTypes` too.

But yeah, I think we can start by migrating the `HTTPHeaders`, and then migrate HTTPRequest/HTTPResponse in a new PR, or wait until they have the `body` also represented.

zunda-pixel:
Hi! @grdsdev I tried to replace HTTPRequest, but It is hard...
Yes, HTTPTypes.HTTPRequest has no body.
I replaced `HTTP Header` and `HTTP Method`

Tests/PostgRESTTests/PostgrestBuilderTests.swift
```diff
@@ -1,23 +0,0 @@
...
-//  PostgrestBuilderTests.swift
-//  Supabase
-//
-//  Created by Guilherme Souza on 20/08/24.
-//
-
-@testable import PostgREST
-import XCTest
-
-final class PostgrestBuilderTests: XCTestCase {
```
grdsdev, line 11: Why was this test removed?

grdsdev (changes requested):

Tests/PostgRESTTests/PostgrestBuilderTests.swift
zunda-pixel: Because This test has only Custom Header Test.
Shouldn't we remove?

grdsdev: This actually tests that when you set custom headers on a builderB that was derived from a builderA, builderA doesn't get affected.

We should keep the test.

Tests/IntegrationTests/Potsgrest/PostgrestResourceEmbeddingTests.swift
```diff
@@ -70,7 +70,7 @@ final class PostgrestResourceEmbeddingTests: XCTestCase {
       .eq("messages.channel_id", value: 1)
       .execute().value as AnyJSON
 
-    assertInlineSnapshot(of: res, as: .json) {
+    assertInlineSnapshot(of: { res }, as: .json) {
```
grdsdev, line 73: 🤔 why this change? didn't get it.

zunda-pixel: Xcode shows error, but this not error...
I thought this is closure.

```swift
public func assertInlineSnapshot<Value>(
  of value: @autoclosure () throws -> Value?,
```

Sources/PostgREST/PostgrestBuilder.swift
```diff
@@ -51,9 +52,9 @@ public class PostgrestBuilder: @unchecked Sendable {
 
   /// Set a HTTP header for the request.
   @discardableResult
-  public func setHeader(name: String, value: String) -> Self {
+  public func setHeader(name: HTTPField.Name, value: String) -> Self {
```
grdsdev, line 55: This is breaking change, lets keep the name as a String and create the HTTPField.Name internally.

zunda-pixel: HTTPField.Name.init?(_ name: String) is optional init.
Can we use force unwrap?

```swift
@discardableResult
public func setHeader(name: String, value: String) -> Self {
  return self.setHeader(
    name: HTTPField.Name(name)!,
    value: value
  )
}
```

grdsdev: yeah, force unwrap here is fine.

zunda-pixel: ok, I. added.

grdsdev (approved):

Issue #596, opened 2024-11-08
Token refresh offline deletes session


stephenth0ma5:
# Bug report

## Describe the bug

- `NSError` that are `URLError` are failing the `RetryableError` coercion causing the session to be deleted when a token is refreshed offline

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Sign a user in
2. Change date ahead to force a refresh
3. Go offline
4. Launch app
5. Session is deleted

Reproducible in this playground as well:
```
/// An error type that can be retried.
protocol RetryableError: Error {
  /// Whether this error instance should be retried or not.
  var shouldRetry: Bool { get }
}

extension URLError: RetryableError {
  var shouldRetry: Bool {
    defaultRetryableURLErrorCodes.contains(code)
  }
}

/// The default set of retryable URL error codes.
let defaultRetryableURLErrorCodes: Set<URLError.Code> = [
  .backgroundSessionInUseByAnotherProcess, .backgroundSessionWasDisconnected,
  .badServerResponse, .callIsActive, .cannotConnectToHost, .cannotFindHost,
  .cannotLoadFromNetwork, .dataNotAllowed, .dnsLookupFailed,
  .downloadDecodingFailedMidStream, .downloadDecodingFailedToComplete,
  .internationalRoamingOff, .networkConnectionLost, .notConnectedToInternet,
  .secureConnectionFailed, .serverCertificateHasBadDate,
  .serverCertificateNotYetValid, .timedOut,
]

/// The default set of retryable HTTP status codes.
let defaultRetryableHTTPStatusCodes: Set<Int> = [
  408, 500, 502, 503, 504,
]

let urlError = URLError(.notConnectedToInternet)

let userInfo: [String: Any] = [
    NSLocalizedDescriptionKey: "The Internet connection appears to be offline.",
    NSURLErrorFailingURLStringErrorKey: "https://awfgscmxufjzehndzbsx.supabase.co/auth/v1/token?grant_type=refresh_token",
    NSURLErrorFailingURLErrorKey: "https://awfgscmxufjzehndzbsx.supabase.co/auth/v1/token?grant_type=refresh_token",
    "_kCFStreamErrorCodeKey": 50,
    "_NSURLErrorNWResolutionReportKey": "Resolved 0 endpoints in 1ms using unknown from cache",
    "_NSURLErrorNWPathKey": "unsatisfied (No network route)",
    "_kCFStreamErrorDomainKey": 1,
[...]

PR #597, opened 2024-11-08
fix(auth): `URLError` coercion for `RetryableError` causing session to be deleted

stephenth0ma5:

Fixes the issue described in https://github.com/supabase/supabase-swift/issues/596

- Only strictly `URLError` are retried or any error that directly conforms to `RetryableError`

## What is the new behavior?
- Any `NSError` or `Error` that conforms to `URLError` and is retryable will be retried along with any error that directly conforms to `RetryableError`

### DIFF
--- a/Sources/Auth/Internal/SessionManager.swift
+++ b/Sources/Auth/Internal/SessionManager.swift
@@ -108,6 +108,8 @@ private actor LiveSessionManager {
             // the same is true for AuthError.
             if let error = error as? URLError, error.shouldRetry {
               throw error
+            } else if let error = error as? any RetryableError, error.shouldRetry {
+              throw error
             } else {
               remove()
               throw error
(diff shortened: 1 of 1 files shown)


Sources/Auth/Internal/SessionManager.swift
```diff
@@ -106,7 +106,9 @@ private actor LiveSessionManager {
             // Need to do this check here, because not all RetryableError's should be retried.
             // URLError conforms to RetryableError, but only a subset of URLError should be retried,
             // the same is true for AuthError.
-            if let error = error as? any RetryableError, error.shouldRetry {
+            if let error = error as? URLError, error.shouldRetry {
+              throw error
+            } else if let error = error as? any RetryableError, error.shouldRetry {
```
grdsdev, line 111: Hi, can you add a comment in the code on why this is needed?

grdsdev (approved):

PR #613, opened 2024-11-30
Change to function from closure

zunda-pixel:

Fix eror in Swift 6

```
Static property 'reconnectSteppedBackOff' is not concurrency-safe because non-'Sendable' type '(Int) -> TimeInterval' (aka '(Int) -> Double') may have shared mutable state; this is an error in the Swift 6 language mode
```


grdsdev:
The library doesn’t support Swift 6 mode yet. The specific part of the code is deprecated and will be removed in the next major version. Therefore, it’s not worth modifying it at the moment.

PR #633, opened 2025-01-08
refactor: use Clock protocol when available

grdsdev:

refactor

Current code uses `Task.sleep` for waiting for time to pass, this isn't ideal when testing, since in tests we need to actually wait for time to pass.

Example, for testing a code that has a `Task.sleep(nanoseconds: NSEC_PER_SEC * 10)`, we need to actually wait for 10 seconds in test, which make tests take longer than it should.

## What is the new behavior?

Latest Swift offers a new protocol named `Clock` which solves this issue since there is a `TestClock` implementation which can be used on tests to "fake time to pass", unfortunately `Clock` is available only on newer OS versions, and Supabase support older platform for now.

So in this PR I had to add an intermediate `_Clock` protocol to make it possible to use `Clock` when available and fallback to the old `Task.sleep` on old platforms. This works fine since we always run tests on latest OS versions.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -28,6 +28,7 @@ let package = Package(
     .package(url: "https://github.com/pointfreeco/swift-custom-dump", from: "1.3.2"),
     .package(url: "https://github.com/pointfreeco/swift-snapshot-testing", from: "1.17.2"),
     .package(url: "https://github.com/pointfreeco/xctest-dynamic-overlay", from: "1.2.2"),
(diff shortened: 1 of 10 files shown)


dshukertjr:
It seems like one of the tests is failing.

grdsdev:
@dshukertjr I just re-run the test and it passed, not sure why it failed, it wasn't not even a realtime test 🤔

dshukertjr (approved):

PR #647, opened 2025-01-24
test(functions): increase code coverage

grdsdev:

I first started doing it in https://github.com/supabase/supabase-swift/pull/645 but it was becoming a quite big PR to review, so I decided to break it into smaller PRs.

This is the first PR of many for increasing code coverage for the library, focusing on `Functions`

### DIFF
--- a/Sources/Functions/FunctionsClient.swift
+++ b/Sources/Functions/FunctionsClient.swift
@@ -43,15 +43,31 @@ public final class FunctionsClient: Sendable {
   ///   - region: The Region to invoke the functions in.
   ///   - logger: SupabaseLogger instance to use.
   ///   - fetch: The fetch handler used to make requests. (Default: URLSession.shared.data(for:))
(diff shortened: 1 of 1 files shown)


Sources/TestHelpers/URLRequestSnapshot.swift
```diff
@@ -0,0 +1,110 @@
...
+    // Records:
+    //
+    // ```
+    // curl \
+    //   --request POST \
+    //   --header "Accept: text/html" \
+    //   --data 'pricing[billing]=monthly&pricing[lane]=individual' \
+    //   "https://www.pointfree.co/subscribe"
+    // ```
+    package static let _curl = SimplySnapshotting.lines.pullback { (request: URLRequest) in
```
grdsdev, line 33: As discussed in https://github.com/pointfreeco/swift-snapshot-testing/discussions/945, I had to write our own snapshot strategy that uses both `httpBody` and `httpBodyStream`, as this won't be merged upstream.

dshukertjr (approved):

PR #649, opened 2025-01-25
chore: fix Swift Package Manager dependency instruction

jmshrv:

The code snippet in README.md gave me the error "product 'Supabase' required by package '<package>' target '<target>' not found. Did you mean '.product(name: "Supabase", package: "supabase-swift")'?". Applying the suggested fix worked, so I thought I'd update the README to do this directly :)

When just declaring Supabase as a dependency, Xcode fails to find the relevant package.

## What is the new behavior?

With this change, Xcode can properly find the Supabase package.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,9 @@ let package = Package(
     targets: [
         .target(
             name: "YourTargetName",
-            dependencies: ["Supabase"] // Add as a dependency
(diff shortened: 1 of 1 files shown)


grdsdev (approved):
Thanks

PR #669, opened 2025-02-19
[RealTimeChannelV2] Add filters instead of regular `String`

LucasAbijmil:
This PR introduces static filters for `RealtimeChannelV2` instead of using a regular `String` (see my [Feature Request](https://github.com/orgs/supabase/discussions/33619#discussioncomment-12237630))

A `String` passed as a parameter by the client (error prone)

## What is the new behavior?
Use of SDK-defined filters, similar in syntax to a regular query.

### DIFF
--- a/Sources/Realtime/RealtimeChannel+AsyncAwait.swift
+++ b/Sources/Realtime/RealtimeChannel+AsyncAwait.swift
@@ -6,6 +6,36 @@
 //
 
 import Foundation
+import PostgREST
+
+public enum RealtimeChannelV2Filter {
+  case eq(column: String, value: any URLQueryRepresentable)
+  case neq(column: String, value: any URLQueryRepresentable)
+  case gt(column: String, value: any URLQueryRepresentable)
+  case gte(column: String, value: any URLQueryRepresentable)
+  case lt(column: String, value: any URLQueryRepresentable)
+  case lte(column: String, value: any URLQueryRepresentable)
+  case `in`(column: String, value: [any URLQueryRepresentable])
+
+  var value: String {
+    switch self {
+    case .eq(let column, let value):
+      return "\(column)=eq.\(value.queryValue)"
+    case .neq(let column, let value):
+      return "\(column)=neq.\(value.queryValue)"
+    case .gt(let column, let value):
+      return "\(column)=gt.\(value.queryValue)"
+    case .gte(let column, let value):
+      return "\(column)=gte.\(value.queryValue)"
+    case .lt(let column, let value):
+      return "\(column)=lt.\(value.queryValue)"
+    case .lte(let column, let value):
+      return "\(column)=lte.\(value.queryValue)"
+    case .in(let column, let value):
+      return "\(column)=in.(\(value.map(\.queryValue).joined(separator: ",")))"
+    }
+  }
+}
 
 extension RealtimeChannelV2 {
   /// Listen for clients joining / leaving the channel using presences.
@@ -28,6 +58,22 @@ extension RealtimeChannelV2 {
     _: InsertAction.Type,
     schema: String = "public",
     table: String? = nil,
+    filter: RealtimeChannelV2Filter? = nil
+  ) -> AsyncStream<InsertAction> {
+    postgresChange(event: .insert, schema: schema, table: table, filter: filter?.value)
+      .compactErase()
+  }
+
+  /// Listen for postgres changes in a channel.
+  @available(
+    *,
+     deprecated,
+     message: "Use the new filter syntax instead."
+  )
+  public func postgresChange(
+    _: InsertAction.Type,
+    schema: String = "public",
+    table: String? = nil,
     filter: String? = nil
   ) -> AsyncStream<InsertAction> {
     postgresChange(event: .insert, schema: schema, table: table, filter: filter)
@@ -35,6 +81,22 @@ extension RealtimeChannelV2 {
   }
 
   /// Listen for postgres changes in a channel.
+  public func postgresChange(
+    _: UpdateAction.Type,
+    schema: String = "public",
+    table: String? = nil,
+    filter: RealtimeChannelV2Filter? = nil
+  ) -> AsyncStream<UpdateAction> {
+    postgresChange(event: .update, schema: schema, table: table, filter: filter?.value)
+      .compactErase()
+  }
+
+  /// Listen for postgres changes in a channel.
+  @available(
+    *,
+     deprecated,
+     message: "Use the new filter syntax instead."
+  )
   public func postgresChange(
     _: UpdateAction.Type,
     schema: String = "public",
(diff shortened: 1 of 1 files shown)


Sources/Realtime/RealtimeChannel+AsyncAwait.swift
```diff
@@ -6,6 +6,36 @@
...
+import PostgREST
+
+public enum RealtimeChannelV2Filter {
+  case eq(column: String, value: any URLQueryRepresentable)
+  case neq(column: String, value: any URLQueryRepresentable)
+  case gt(column: String, value: any URLQueryRepresentable)
+  case gte(column: String, value: any URLQueryRepresentable)
+  case lt(column: String, value: any URLQueryRepresentable)
+  case lte(column: String, value: any URLQueryRepresentable)
+  case `in`(column: String, value: [any URLQueryRepresentable])
```
LucasAbijmil, line 18: @grdsdev I don't know if I should use `URLQueryRepresentable` here – The benefit is that we already have all the values associated with many types.

Sources/Realtime/RealtimeChannel+AsyncAwait.swift
```diff
@@ -6,6 +6,36 @@
...
+      return "\(column)=gte.\(value.queryValue)"
+    case let .lt(column, value):
+      return "\(column)=lt.\(value.queryValue)"
+    case let .lte(column, value):
+      return "\(column)=lte.\(value.queryValue)"
+    case let .in(column, values):
+      return "\(column)=in.(\(values.map(\.queryValue).joined(separator: ",")))"
+    }
+  }
+}
```
LucasAbijmil, line 38: @grdsdev I don't know if I should use `URLQueryRepresentable` here – The benefit is that we already have all the values associated with many types.

grdsdev: Hey @LucasAbijmil I don't think the `URLQueryRepresentable` is a good option here since it is implemented by types that aren't supported in the realtime filter, for example `JSONObject`, and Array (which is a different representation).

I'd rather accept only `String` values, and the caller should transform to string

Sources/Realtime/RealtimeChannel+AsyncAwait.swift
```diff
@@ -6,6 +6,36 @@
 //
 
 import Foundation
+import PostgREST
```
grdsdev, line 9: `Realtime` shouldn't depend on `PostgREST`

Sources/Realtime/RealtimeChannel+AsyncAwait.swift
```diff
@@ -24,6 +54,22 @@ extension RealtimeChannelV2 {
...
+    postgresChange(event: .insert, schema: schema, table: table, filter: filter?.value)
+      .compactErase()
+  }
+
+  /// Listen for postgres changes in a channel.
+  @available(
+    *,
+     deprecated,
+     message: "Use the new filter syntax instead."
+  )
```
grdsdev, line 72: I think you'll have to mark all deprecated method also as [`@_disfavoredOverload`](https://www.fivestars.blog/articles/disfavoredOverload/)

grdsdev (changes requested):
Thanks for the work here, just a few change requests.

We'd also need some tests for the new filter type.

Sources/Realtime/RealtimeChannel+AsyncAwait.swift
grdsdev: I'm changing the URLQueryRepresentable to make it specific to Postgrest filter values: https://github.com/supabase/supabase-swift/pull/670

Sources/Realtime/RealtimeChannel+AsyncAwait.swift
LucasAbijmil: Indeed

LucasAbijmil:
@grdsdev Let me know what you think of my update, and I'll finalize the tests based on your feedback

Sources/Realtime/RealtimeChannel+AsyncAwait.swift
LucasAbijmil: I based my `RealtimeFilter` on your changes :)

Sources/Realtime/RealtimeFilterValue.swift
```diff
@@ -0,0 +1,47 @@
...
+//  RealtimeFilterValue.swift
+//  Supabase
+//
+//  Created by Lucas Abijmil on 19/02/2025.
+//
+
+import Foundation
+
+/// A value that can be used to filter Realtime changes in a channel.
+public protocol RealtimeFilterValue {
```
grdsdev, line 11:
```suggestion
public protocol RealtimePostgresFilterValue {
```

Sources/Realtime/RealtimeFilter.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  RealtimeFilter.swift
+//  Supabase
+//
+//  Created by Lucas Abijmil on 19/02/2025.
+//
+
+/// A filter that can be used in Realtime.
+public enum RealtimeFilter {
```
grdsdev, line 9:
```suggestion
public enum RealtimePostgresFilter {
```

Sources/Realtime/RealtimeFilterValue.swift
```diff
@@ -0,0 +1,47 @@
...
+    formatter.formatOptions = [.withInternetDateTime, .withFractionalSeconds]
+    return formatter.string(from: self)
+  }
+}
+
+extension Array: RealtimeFilterValue where Element: RealtimeFilterValue {
+  public var rawValue: String {
+    map(\.rawValue).joined(separator: ",")
+  }
+}
```
grdsdev, line 47: Lets remove the Array conformance, since the `in` filter already accepts a `[any RealtimePostgresFilterValue]`, we don't want to allow Arrays in other filters.

```suggestion
```

grdsdev (changes requested):
You're on the right track, just a few more changes.

Sources/Realtime/RealtimeFilterValue.swift
LucasAbijmil: Yes you're right

LucasAbijmil:
> You're on the right track, just a few more changes.

I hope you'll enjoy the changes :)

AronnK:
Hey! So I'm trying to have 2 filters in my code, one .eq to a row and one change in a specific column (corresponding to that row)
So is it possible to do it now by using AND or some other means?

grdsdev:
Hi @AronnK  multiple filters aren't supported yet, follow the thread https://github.com/supabase/realtime-js/issues/97 to know more.

AronnK:
> Hi @AronnK multiple filters aren't supported yet, follow the thread [supabase/realtime-js#97](https://github.com/supabase/realtime-js/issues/97) to know more.

Hey @grdsdev, thanks for the information!

PR #721, opened 2025-05-28
fix(realtime): make Push associated to MainActor

grdsdev:

Bug fix

`Push` is currently implemented as an `Actor`.

## What is the new behavior?

Refactor `Push` to be a class attached to the `MainActor`, which makes it easier to debug and reason about behavior.

## Additional context

### DIFF
--- a/Tests/RealtimeTests/_PushTests.swift
+++ b/Tests/RealtimeTests/_PushTests.swift
@@ -12,83 +12,85 @@ import XCTest
 
 @testable import Realtime
 
-@MainActor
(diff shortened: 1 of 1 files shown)


Sources/Realtime/RealtimeChannelV2.swift
```diff
@@ -81,6 +83,7 @@ public final class RealtimeChannelV2: Sendable {
   }
 
   /// Subscribes to the channel
+  @MainActor
```
grdsdev, line 86: This isn't a breaking change since the method is `async`, the call site already have to call this method with an `await` independent of the running actor.

dshukertjr (approved):

Issue #727, opened 2025-06-05
Edge function timeout 60 seconds


Magnuti:
# Bug report

- [x] I confirm this is a bug with Supabase, not with my own application. 
- [x] I confirm I have searched the [Docs](https://docs.supabase.com), GitHub [Discussions](https://github.com/supabase/supabase/discussions), and [Discord](https://discord.supabase.com). 

## Describe the bug

Edge functions timeouts after 60 seconds

```
Error Domain=NSURLErrorDomain Code=-1001 "Timeout for the request." UserInfo={_kCFStreamErrorCodeKey=-2102, NSUnderlyingError=0x128c9a4f0 {Error Domain=kCFErrorDomainCFNetwork Code=-1001 "(null)" UserInfo={_kCFStreamErrorCodeKey=-2102, _kCFStreamErrorDomainKey=4}}, _NSURLErrorFailingURLSessionTaskErrorKey=LocalDataTask <3922FB75-06BE-48FF-96B3-F79C4227EAFA>.<24>, _NSURLErrorRelatedURLSessionTaskErrorKey=(
    "LocalDataTask <3922FB75-06BE-48FF-96B3-F79C4227EAFA>.<24>"
), NSLocalizedDescription=Timeout for the request., NSErrorFailingURLStringKey=http://10.152.9.218:5432/functions/v1/my-function, NSErrorFailingURLKey=http://10.152.9.218:5432/functions/v1/my-function, _kCFStreamErrorDomainKey=4}
```

## To Reproduce

Steps to reproduce the behavior, please provide code snippets or a repository:

1. Create an edge function with a timeout of 60 seconds or more `await new Promise((resolve) => setTimeout(resolve, 60 * 1000))`
2. Invoke the edge function `try await supabaseClient.functions.invoke("my-function")`
3. Wait for response
4. Get timeout

## Expected behavior

Should not time out after 60 seconds, but after 150 seconds as specified here https://supabase.com/docs/guides/functions/limits

## System information

- supabase-swift 2.26.0
- Supabase: 2.23.4
- deno 2.3.1
- Node v22.15.0

## Additional context

How can we increase the timeout to 150 seconds?

grdsdev:
Hi @Magnuti `60 seconds` is the default timeout of `URLSession`, will increase it to 150s for edge functions, thanks for reporting this.

PR #728, opened 2025-06-05
fix(functions): set request timeout to 150 seconds when invoking functions

grdsdev:

Bug fix
Close https://github.com/supabase/supabase-swift/issues/727

No request timeout defined, which uses default `URLSession` timeout which is 60 seconds.

## What is the new behavior?

Allow to specify a custom value for timeout a per-request basis, (internally, not configurable by the user)
Set all invoke functions requests timeout to 150 seconds

## Additional context

https://supabase.com/docs/guides/functions/limits

### DIFF
--- a/Sources/Functions/FunctionsClient.swift
+++ b/Sources/Functions/FunctionsClient.swift
@@ -15,6 +15,11 @@ public final class FunctionsClient: Sendable {
     Data, URLResponse
   )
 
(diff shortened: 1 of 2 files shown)


grdsdev:
@dshukertjr not sure if this happens in Flutter too, does HTTPClient has a default timeout for the requests? If that is the case, this will need to be fixed on flutter too.

dshukertjr (approved):

PR #749, opened 2025-07-21
fix(realtime): wildcard event not handled in onBroadcast callback

wzulfikar:

Bug fix

Using "*" as event arg e.g. `onBroadcast(event: "*")` won't work. The callback is never triggered because the event never matched.

```swift
let subscription = = channel.onBroadcast(event: "*") { message in
    print("new message:", message) // This never gets logged
}
```

I found it because I thought I had a bug in my app code, but once I enabled global logging, I got the log from `RealtimeClientV2.onMessage` which means the event was actually sent and received, but the callback isn't triggered.

image

## What is the new behavior?

Using "*" as event arg should work now.

## Additional context

The behavior is documented [here](https://supabase.com/docs/guides/realtime/broadcast?queryGroups=language&language=swift#receiving-broadcast-messages). It's also the same behavior (ie. treating "*" as wildcard) in other languages.

image

### DIFF
--- a/Sources/Realtime/CallbackManager.swift
+++ b/Sources/Realtime/CallbackManager.swift
@@ -124,7 +124,7 @@ final class CallbackManager: Sendable {
       }
       return nil
     }
-    let callbacks = broadcastCallbacks.filter { $0.event == event }
+    let callbacks = broadcastCallbacks.filter { $0.event == "*" || $0.event.lowercased() == event.lowercased() }
     callbacks.forEach { $0.callback(json) }
   }
 
--- a/Tests/RealtimeTests/CallbackManagerTests.swift
+++ b/Tests/RealtimeTests/CallbackManagerTests.swift
@@ -203,7 +203,7 @@ final class CallbackManagerTests: XCTestCase {
     callbackManager.triggerBroadcast(event: "NEW_USER", json: jsonObject)
     XCTAssertEqual(caseInsensitiveMessage.value, jsonObject)
 
-    // Match wildcard event
+    // Match any events with wildcard
     let wildcardReceivedMessage = LockIsolated<JSONObject?>(nil)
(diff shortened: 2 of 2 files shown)


Sources/Realtime/CallbackManager.swift
```diff
@@ -124,7 +124,7 @@ final class CallbackManager: Sendable {
       }
       return nil
     }
-    let callbacks = broadcastCallbacks.filter { $0.event == event }
+    let callbacks = broadcastCallbacks.filter { $0.event == "*" || $0.event.lowercased() == event.lowercased() }
```
wzulfikar, line 127: The main change for the fix.

We handle "*" similarly for the non-callback format:

https://github.com/supabase/supabase-swift/blob/84f0f14a5cc59d965f874c77e7aee16332e6c6e7/Sources/Realtime/Deprecated/RealtimeChannel.swift#L928-L934

Also checked other repos (ts + py) to make sure we don't do anything further than matching "*".

[supabase/realtime-py](https://github.com/supabase/realtime-py/blob/6e4154e02b1bedc0fae70183abbf534358c8b36a/realtime/_async/channel.py#L571):

image

[supabase/realtime-js](https://github.com/supabase/realtime-js/blob/d3026be84cd2ed2ac1b22deda4bbfc18e0226d86/src/RealtimeChannel.ts#L663-L665):

image

---

Tangential: the case insensitive check doesn't seem right because it only `.lowercased()` the message payload type but not the `bindEvent`. Didn't change anything there, just pointing it out in case it's not intentional.

https://github.com/supabase/supabase-swift/blob/84f0f14a5cc59d965f874c77e7aee16332e6c6e7/Sources/Realtime/Deprecated/RealtimeChannel.swift#L928-L934

grdsdev:
Great catch @wzulfikar thanks.

PR #757, opened 2025-07-29
feat: add OSLogSupabaseLogger type

grdsdev:

Improvement

User needs to manually implement a `SupabaseLogger` which uses a `OSLog`.

## What is the new behavior?

We now provide a `SupabaseLogger` implementation that uses `OSLog`, since this is probably the most used logger across Apple projects.

### DIFF
--- a/Sources/Helpers/Logger/OSLogSupabaseLogger.swift
+++ b/Sources/Helpers/Logger/OSLogSupabaseLogger.swift
@@ -1,0 +1,53 @@
+import Foundation
+
+#if canImport(OSLog)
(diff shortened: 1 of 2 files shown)


dshukertjr (approved):

arbyruns:
Just learning about this. Is there an advantage of using this over implementing logger yourself?

grdsdev:
@arbyruns this is only a convenience logger implementation, if you use OSLog as the logger backend, then it makes sense to use this instead of reimplementing your own.

If you use a different logging backend, then you implement your own.

PR #760, opened 2025-07-31
fix(realtime): make Apikey required

grdsdev:

Bug fix

Realtime client can be initialized without an apikey

## What is the new behavior?

Realtime client now requires apikey to be present

## Additional context

https://github.com/supabase/realtime-js/pull/502


Sources/Realtime/RealtimeChannelV2.swift
```diff
@@ -499,7 +499,7 @@ public final class RealtimeChannelV2: Sendable, RealtimeChannelProtocol {
       )
       Task {
         await unsubscribe()
-        await subscribe()
+        try? await subscribeWithError()
```
grdsdev, line 502: unrelated change, but it just fixes a deprecation warning

PR #844, opened 2025-11-03
fix(auth): replace trait with runtime configuration flag

grdsdev:
## Summary

Replaces the `EmitLocalSessionAsInitialSession` trait with a runtime configuration flag to resolve compatibility issues with Xcode projects and align with Apple's trait guidelines.

## Problem

The current trait-based approach has several issues:
1. **Xcode incompatibility**: iOS projects cannot enable traits (no Package.swift, no UI/CLI support)
2. **Violates Apple's guidelines**: Traits should be strictly additive and not alter API behavior
3. **Limited discoverability**: Users in Xcode won't know this option exists

## Solution

Implemented a runtime configuration flag `emitLocalSessionAsInitialSession: Bool` in `AuthClient.Configuration`:
- Defaults to `false` for backward compatibility
- Can be set at initialization without recompilation
- Works in all project types (Xcode, SPM)
- Discoverable through autocomplete and documentation

## Changes

- ✅ Added `emitLocalSessionAsInitialSession: Bool` property to `AuthClient.Configuration`
- ✅ Replaced conditional compilation (`#if EmitLocalSessionAsInitialSession`) with runtime checks
- ✅ Updated deprecation warning to reference the new configuration option
- ✅ Added separate tests for both behaviors (old and new)
- ✅ Removed `EmitLocalSessionAsInitialSession` trait from Package@swift-6.1.swift
- ✅ All 78 AuthClientTests passing

## Migration

Users who want the new behavior can now set:

```swift
let client = AuthClient(
  url: url,
  headers: headers,
  localStorage: storage,
  emitLocalSessionAsInitialSession: true  // Opt-in to new behavior
)
```

Or using configuration:

```swift
let config = AuthClient.Configuration(
  url: url,
  // ... other parameters
  emitLocalSessionAsInitialSession: true
)
let client = AuthClient(configuration: config)
```

## Test Plan

- [x] Build succeeds (`swift build`)
- [x] All AuthClientTests pass (78/78)
- [x] Tests cover both old behavior (default) and new behavior (with flag enabled)
- [x] No breaking changes for existing users

## Breaking Changes

None. The default behavior remains unchanged (`false`). This will change to `true` in the next major release (v3.0.0).

### DIFF
--- a/Sources/Supabase/SupabaseClient.swift
+++ b/Sources/Supabase/SupabaseClient.swift
@@ -181,7 +181,8 @@ public final class SupabaseClient: Sendable {
         // DON'T use `fetchWithAuth` method within the AuthClient as it may cause a deadlock.
         try await options.global.session.data(for: $0)
       },
-      autoRefreshToken: options.auth.autoRefreshToken
+      autoRefreshToken: options.auth.autoRefreshToken,
+      emitLocalSessionAsInitialSession: options.auth.emitLocalSessionAsInitialSession
     )
 
     _realtime = UncheckedSendable(
--- a/Sources/Supabase/Types.swift
+++ b/Sources/Supabase/Types.swift
@@ -57,6 +57,13 @@ public struct SupabaseClientOptions: Sendable {
     /// Set to `true` if you want to automatically refresh the token before expiring.
     public let autoRefreshToken: Bool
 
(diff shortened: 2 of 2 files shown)


vojtabohm:
Seems to be working as intended. Going to production with it. Thanks!

mandarini (approved):

PR #855, opened 2025-11-20
fix(realtime): extract ConnectionManager actor and fix connection lifecycle races

grdsdev:
## Summary

Extracts WebSocket connection lifecycle out of `RealtimeClientV2` into a new `ConnectionManager` actor, driven by an explicit state machine.

## Motivation

Connection state in `RealtimeClientV2` was spread across several fields and ad-hoc tasks — current socket, in-flight connect task, reconnect task, heartbeat task, status subject — all coordinated through `LockIsolated` mutable state. This made it hard to reason about transitions, hard to test reconnection in isolation, and easy to introduce races when adding features.

## Bugs fixed

1. **Lost `phx_join` reply after connect.** `WebSocket.events` lazily installs `onEvent` only when the stream is first iterated. The old code read `conn.events` inside the message `Task` body, so the server's `phx_reply` could arrive before the listener attached and get dropped — channels stuck in `.subscribing`. Fixed by hoisting `let stream = conn.events` out of the task.
2. **`status` race on `connect()` return.** `status` was updated asynchronously by the observer, so code running right after `await connect()` (e.g. `RealtimeChannelV2._subscribe` checking `socket.status == .connected`) could see stale `.connecting` and bail. Fixed by yielding status synchronously in `connect()` / `disconnect()`.
3. **Duplicate / missing status transitions.** The old path could emit `.connecting` twice (once for `.connecting`, once for `.reconnecting`) or skip `.disconnected` between `.connected` → `.reconnecting`. Fixed with dedup plus an explicit `.disconnected` emission before reconnect.
4. **Concurrent `connect()` calls racing.** Two overlapping `connect()` invocations could each spawn a transport, leaking sockets. The actor's `.connecting` case now awaits the in-flight task instead of starting a new one.
5. **`disconnect()` not cancelling an in-flight connection.** The connect task could complete after `disconnect()` returned and re-enter `.connected`. `disconnect` now cancels the `.connecting` / `.reconnecting` task.
6. **Double-close on remote-initiated close.** `handleClose` used to send another close frame to an already-closed socket. Now it only updates state.
7. **Reconnect not testable / fired under cancellation.** Old reconnect used `Task.sleep`, ignoring the injected `_clock`, and `CancellationError` could trigger a reconnect loop. Now uses `_clock.sleep(for:)` and skips reconnect for `CancellationError`.
8. **Task leaks across reconnects.** `messageTask` / `heartbeatTask` weren't consistently cancelled before being replaced. `handleConnected` now cancels-then-replaces both.

### `ConnectionManager` (new actor)

Owns connection lifecycle behind an explicit state:

```
.disconnected
.connecting(Task)
.connected(WebSocket)
.reconnecting(Task, reason)
```

Responsibilities:
- `connect()` — opens a socket; concurrent callers await the in-flight task instead of racing.
- `disconnect(reason:)` — cancels in-flight attempts or closes a live socket; updates state to `.disconnected`.
- `handleError(_:)` — closes the socket and schedules a reconnect on `_clock` (so tests can advance time).
- `handleClose(code:reason:)` — remote-initiated close; updates state without sending an extra close frame.
- `stateChanges` — `nonisolated AsyncStream<State>` for observers (no actor hop to subscribe).

### `RealtimeClientV2`

- Delegates the socket lifecycle to `ConnectionManager` and observes it through a single long-lived task (weak-self captured, stored for cancellation).
- `connect()` drives post-connect setup synchronously (message listening, heartbeat, send-buffer flush) so callers can rely on state being ready on return. The observer only drives setup on automatic reconnects (`.reconnecting` → `.connected`).
- Status transitions are yielded synchronously from `connect()` / `disconnect()` with dedup against the last value. The observer only emits `.connected` on auto-reconnect and `.connecting` on `.reconnecting`, so subscribers never see duplicate transitions.
- `listenForMessages` installs `WebSocket.onEvent` synchronously (by reading `conn.events` before spawning the message task) so replies to `phx_join` don't land before the listener is attached.

### Tests

- New `ConnectionManagerTests` exercises the state machine directly: connect/connected transitions, dedup of concurrent connects, disconnect cancelling an in-flight attempt, reconnect after error, `handleClose` not double-closing.
- `RealtimeTests` updated where internal call sites moved; no behavioural changes required.

## Public API

No breaking changes. All public signatures are preserved — `disconnect()` stays synchronous (the actor call is dispatched on a detached `Task`).

## Test results

147/147 Realtime tests pass locally (`swift test --filter RealtimeTests`).

### DIFF
--- a/Sources/Realtime/ConnectionManager.swift
+++ b/Sources/Realtime/ConnectionManager.swift
@@ -1,0 +1,180 @@
+//
+//  ConnectionManager.swift
+//  Supabase
+//
+//  Created by Guilherme Souza on 19/11/25.
+//
+
+import ConcurrencyExtras
+import Foundation
+
+actor ConnectionManager {
+  enum State {
+    case disconnected
+    case connecting(Task<Void, any Error>)
+    case connected(any WebSocket)
+    case reconnecting(Task<Void, any Error>, reason: String)
+  }
+
+  private let (stateStream, stateContinuation) = AsyncStream<State>.makeStream()
+  private(set) var state: State = .disconnected
+
+  private let transport: WebSocketTransport
+  private let url: URL
+  private let headers: [String: String]
+  private let reconnectDelay: TimeInterval
+  private let logger: (any SupabaseLogger)?
+
+  /// Get current connection if connected, nil otherwise.
+  var connection: (any WebSocket)? {
+    if case .connected(let conn) = state {
+      return conn
+    }
+    return nil
+  }
+
+  nonisolated var stateChanges: AsyncStream<State> { stateStream }
+
+  init(
+    transport: @escaping WebSocketTransport,
+    url: URL,
+    headers: [String: String],
+    reconnectDelay: TimeInterval,
+    logger: (any SupabaseLogger)?
+  ) {
+    self.transport = transport
+    self.url = url
+    self.headers = headers
+    self.reconnectDelay = reconnectDelay
+    self.logger = logger
+  }
+
+  @discardableResult
+  func connect() async throws -> any WebSocket {
+    logger?.debug("current state: \(state)")
+
+    switch state {
+    case .connected(let conn):
+      logger?.debug("Already connected")
+      return conn
+
+    case .connecting(let task):
+      logger?.debug("Connection already in progress, waiting...")
+      try await task.value
+      // After waiting, get the connection from state
+      guard case .connected(let conn) = state else {
+        throw WebSocketError.connection(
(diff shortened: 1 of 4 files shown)


Sources/Realtime/ConnectionManager.swift
```diff
@@ -0,0 +1,180 @@
+//
+//  ConnectionManager.swift
+//  Supabase
+//
+//  Created by Guilherme Souza on 19/11/25.
```
mandarini: is this needed? i think date conflicts with git blame?

mandarini (approved):

PR #878, opened 2026-01-03
fix(functions): sync auth headers on auth state change

### LINKED ISSUE #877: `_invokeWithStreamedResponse` sends stale auth token after sign in
`FunctionsClient` auth headers are not updated when auth state changes. While `realtime` and `realtimeV2` receive `setAuth()` calls in `handleTokenChanged()`, `functions` is missing.

This causes `_invokeWithStreamedResponse` to send the anon key instead of the user's access token, since it reads directly from `mutableState.headers` which never gets updated on sign in/out/token refresh.

7ttp:
## Summary
FunctionsClient auth headers are not updated when auth state changes.

## Problem
While `realtime` and `realtimeV2` receive `setAuth()` calls in `handleTokenChanged()`, `functions` was missing. This causes `_invokeWithStreamedResponse` to send the anon key instead of the user's access token.

## Solution
Add `mutableState.functions?.setAuth(token: accessToken)` in `handleTokenChanged()` to sync auth headers when user signs in/out or token refreshes.

## Related
Closes #877

### DIFF
--- a/Sources/Supabase/SupabaseClient.swift
+++ b/Sources/Supabase/SupabaseClient.swift
@@ -388,9 +388,11 @@ public final class SupabaseClient: Sendable {
       return nil
     }
 
-    realtime.setAuth(accessToken)
-    await realtimeV2.setAuth(accessToken)
-    functions.setAuth(token: accessToken)
+    if let accessToken {
+      functions.setAuth(token: accessToken)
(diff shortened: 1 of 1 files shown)


7ttp:
@grdsdev

grdsdev (approved):

PR #959, opened 2026-04-16
chore(ci): improve coverage and parallelism

grdsdev:
## Summary

- **New platform coverage**: tvOS, watchOS, and visionOS are now build-verified in the `macos` matrix. These platforms are declared supported in `Package.swift` but were never exercised in CI.
- **Fewer redundant jobs**: Build-only jobs (`command=""`) previously ran both Debug and Release; Debug is now skipped (`skip_debug`) since test jobs already compile in Debug. Removes 2 jobs on latest, 3 on legacy.
- **macOS SPM tests**: The `spm` job now runs `swift test --skip IntegrationTests` on debug, closing the gap where only Linux ran `swift test`.
- **Parallel `examples`**: Converted from three sequential `xcodebuild` calls in one job to a `scheme` matrix (`Examples`, `SlackClone`, `UserManagement`) running in parallel with per-scheme derived-data caches.
- **Removed serial bottleneck**: `examples` and `docs` no longer declare `needs: [macos]` — they start immediately alongside all other jobs, eliminating a ~15–20 min wait from the critical path.
- **Shared SPM cache**: `integration-tests` now uses the same cache key as `linux` (`${{ runner.os }}-spm-`) since both target `ubuntu-latest` and build the same package.
- **Cache before Xcode setup in `macos-legacy`**: Cache restore is now the first step after checkout, before `xcode-select` and `simctl list`.

## Test plan

- [ ] Verify `macos` matrix expands correctly: test × {IOS, MACOS} + build × {IOS, MACOS, TVOS, WATCHOS, VISIONOS}
- [ ] Confirm `macos-legacy` build-only jobs skip Debug and run Release only
- [ ] Confirm `spm` debug job runs `swift test` after `swift build`
- [ ] Confirm `examples` spawns 3 parallel jobs
- [ ] Confirm `docs` and `examples` start without waiting for `macos`
- [ ] Confirm `ci-success` still gates on all expected jobs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,11 +49,18 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
+        # command=test: run tests (Debug only, Release skipped via skip_release)
+        # command="": build only (Release only, Debug skipped via skip_debug)
         command: [test, ""]
-        platform: [IOS, MACOS]
-        xcode: ["26.4"]
+        platform: [IOS, MACOS, TVOS, WATCHOS, VISIONOS]
         include:
           - { command: test, skip_release: 1 }
+          - { command: "", skip_debug: 1 }
+        exclude:
+          # Only build (no test) for non-simulator-friendly platforms
+          - { command: test, platform: TVOS }
+          - { command: test, platform: WATCHOS }
+          - { command: test, platform: VISIONOS }
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Cache derived data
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -49,11 +49,14 @@ jobs:
     timeout-minutes: 30
     strategy:
       matrix:
+        # command=test: run tests (Debug only, Release skipped via skip_release)
+        # command="": build only (Release only, Debug skipped via skip_debug)
```
mandarini: are these needed?

mandarini (approved):

.github/workflows/ci.yml
grdsdev: Removed, good call.

PR #967, opened 2026-04-22
feat(realtime): handle app lifecycle background/foreground transitions

### LINKED ISSUE #595: [Realtime] handle connection when app changes between background/foreground state
# Enchancement

To have a reliable Realtime experience, the client should handle reestablishing connection and resubscribing to channels when it detects that the app moved back to the foreground state.

grdsdev:
## Summary

Adds automatic foreground-recovery to `RealtimeClientV2`. When enabled, the client observes iOS/macOS lifecycle notifications and — on foregrounding — reconnects and re-joins any existing channels **only if the socket was connected when the app backgrounded**. Clients that were intentionally idle when the app backgrounded are left alone.

The client intentionally does **not** disconnect on backgrounding. Short background/foreground cycles (Control Centre, notification pulls) keep the connection alive without churn. If the OS tears the socket down during suspension, the existing error-driven reconnect logic plus the foreground handler recover cleanly.

Closes [#595](https://github.com/supabase/supabase-swift/issues/595).

### Core

- **`Sources/Realtime/RealtimeLifecycleManager.swift`** — now observes **both** foreground and background notifications.
  - Background (`didEnterBackground` / `didResignActive`): calls `handleAppBackground()` synchronously to snapshot connection state.
  - Foreground (`willEnterForeground` / `willBecomeActive`): calls `handleAppForeground()` on a Task.
- **`Sources/Realtime/Types.swift`** — adds `RealtimeClientOptions.handleAppLifecycle` (default `true` on iOS/tvOS/visionOS/macOS, `false` elsewhere).
- **`Sources/Realtime/RealtimeClientV2.swift`**:
  - Adds `wasConnectedBeforeBackground: Bool` to `MutableState`.
  - New `handleAppBackground()`: captures whether the socket was `.connected` at background time.
  - `handleAppForeground()`: no-op unless `wasConnectedBeforeBackground` was `true` **and** the socket is now disconnected; otherwise `connect()` then `rejoinChannels()`.
  - Extracts `rejoinChannels()` as `async` so foreground reconnection can await channel resubscription.

### Tests

- **`Tests/RealtimeTests/RealtimeLifecycleTests.swift`** — 7 tests, all passing:
  1. `testHandleAppForegroundWhileConnectedIsNoOp` — no churn when already connected
  2. `testHandleAppForegroundWithoutPriorBackgroundIsNoOp` — no connection opened if background was never signalled
  3. `testHandleAppForegroundDoesNotConnectIfNotConnectedBeforeBackground` — idle client stays idle
  4. `testHandleAppForegroundReconnectsWhenBackgroundedWhileConnected` — reconnects after background→disconnect
  5. `testHandleAppForegroundResubscribesChannelsWhenBackgroundedWhileConnected` — channels re-join after recovery
  6. `testHandleAppLifecycleFalseDoesNotInstallLifecycleManager`
  7. `testHandleAppLifecycleTrueInstallsLifecycleManager`
- Removed `withMainSerialExecutor` override (caused deadlock with `withThrowingTaskGroup` inside `subscribeWithError`).
- Fixed phx_join auto-reply to mirror the incoming `ref` dynamically (ref counter resets after disconnect).

### Documentation / Example

- **`docs/realtime-app-lifecycle-spec.md`** — cross-platform spec so supabase-js, supabase-flutter, supabase-kt, etc. can implement the same behaviour consistently.
- **`Examples/Examples/Realtime/AppLifecycleView.swift`** — SwiftUI view for manual lifecycle validation.

## Design notes

- **Why capture state on background?** Without it, `handleAppForeground()` could reconnect an intentionally-idle client. Capturing `wasConnectedBeforeBackground` at background time makes the intent explicit and avoids a spurious connection on every foreground event.
- **Why no proactive disconnect on background?** A round-trip disconnect/reconnect for a quick background cycle is pure churn — the socket usually survives. The foreground handler recovers if the OS actually tore it down.
- **Why `AnyObject?` for the stored lifecycle manager?** The concrete type is only compiled on Apple platforms; `AnyObject?` keeps `MutableState` uniform across platforms.

## Test plan

- [x] `swift test --filter RealtimeLifecycleTests` — 7/7 pass on macOS
- [x] `swift test --filter RealtimeTests` — 179/179 pass (no regressions)
- [x] `swift build` clean
- [ ] Manual validation: iOS app subscribes to a channel, backgrounds, returns to foreground → events resume (reviewer)
- [ ] Manual validation: iOS app with no active channel, backgrounds, foregrounds → no spurious connection attempt (reviewer)

## Risk assessment

- **Breaking changes**: none. `RealtimeClientOptions.init` gains a new optional parameter with a platform-conditional default.
- **Backward compatibility**: maintained. Opt out via `handleAppLifecycle: false`.
- **Performance**: negligible — one additional `NotificationCenter` observer per client.
- **Memory**: lifecycle manager holds a `weak` reference to the client; observers are removed in `deinit`.


kkarlsen06:
This has been a pain point, fantastic to see it handled in the package itself!

mandarini (approved):

Issue #579, opened 2024-10-24
[Realtime] auto-reconnect socket and channels when connection is reestablished


grdsdev:
# Enhancement

To have a reliable Realtime experience, client should handle connection issues by reconnecting to socket and resubscribing to channels and it detects the connection was reestablished.

Use `NWPathMonitor` on Apple platforms.

PR #1003, opened 2026-06-01
fix(realtime): prevent deaf socket from duplicate handleConnected (SDK-959)

### LINKED ISSUE #999: [Bug]: Realtime: first channel subscribe stalls for 50s–7min since v2.44.1 (PR #855)
### Version

2.46.0

### Platform

iOS

### Swift Version

6.2

### What happened?

`channel.subscribeWithError()` stalls for tens of seconds (often minutes) on the first JOIN of a fresh socket, since v2.44.1. A clean bisect points to PR #855 (`fix(realtime): extract ConnectionManager actor and fix connection lifecycle races`, commit `dcbc63d5`) as the introducing change. Only one realtime commit lands between v2.44.0 and v2.44.1 — PR #855.

| Version | Subscribe time on cold connect |
|---|---|
| v2.43.1 | ~2.3 s ✅ |
| v2.44.0 | ~3.1 s ✅ |
| v2.44.1 | ~60 s ❌ |
| v2.45.0 | minutes ❌ |
| v2.46.0 | ~4–7 min ❌ |

The clearest signal in the trace: **heartbeats never ack until at least one channel reaches `.subscribed`.** Before subscribe wins, every heartbeat goes `sent` → `timeout`. After subscribe wins, every heartbeat goes `sent` → `ok` instantly. This holds across multiple sessions.

### Hypothesis (not verified beyond the bisect)

PR #855's stated goal includes "Lost phx_join reply after connect: WebSocket.events lazily installed onEvent only when the stream was first iterated. Reading `conn.events` synchronously before spawning the message task ensures replies aren't dropped."

Empirically the change appears to drop early server-to-client frames (both heartbeat `phx_reply`s and the `phx_join` reply) on the first socket — channel stays in `.subscribing`, `pendingHeartbeatRef` is never cleared, the heartbeat-timeout → `connectionManager.handleError` path recycles the socket every ~50 s, and the in-flight JOIN is killed each cycle. Eventually one cycle's JOIN slips through before its 50 s window expires.

This is a hypothesis from reading the v2.44.0 → v2.44.1 diff — the bisect is the load-bearing evidence.

### Server-side context

Project's `realtime` logs show no tenant init/terminate events overlapping the failing subscribe windows; tenant is healthy and serves heartbeat acks instantly once subscribe wins.

REST queries on the same network are fast and reliable throughout.

### Workaround

Pin to `2.44.0`. Subscribes in ~3 s; no further changes needed.

### Possibly using the SDK wrong?

I might be missing a usage-pattern change that landed alongside PR #855 or in subsequent versions. Specifically:

- I'm calling `subscribeWithError()` directly (in a retry-forever wrapper). Is the recommendation now to use `subscribe()` (no `WithError`) and react to `channel.statusChange` instead, given the lifecycle is now managed by `ConnectionManager` / `ChannelStateManager`?
- I start the `postgresChange` AsyncStream consumers *before* calling `subscribeWithError()` (so they're ready when the channel attaches). Is the recommendation now to do this in a different order?
- I use SDK defaults for `handleAppLifecycle` (true on iOS), `connectOnSubscribe` (true), `disconnectOnEmptyChannelsAfter` (default), `timeoutInterval` (10 s), `heartbeatInterval` (25 s), `maxRetryAttempts` (5). Is there a combination that's expected to be configured explicitly for cold-start reliability now?
- Are `postgresChange` bindings on a non-`public` schema (`library` in our case) treated differently in the new state machine?

Happy to be told this is a usage problem rather than a regression — but the bisect is the part I can't explain away on the usage hypothesis (v2.44.0 with the same call pattern works in ~3 s).

### Steps to Reproduce

1. Create a `SupabaseClient` with default `RealtimeClientOptions`.
2. Open one `RealtimeChannelV2`: `let channel = client.channel("library-realtime")`.
3. Bind two `postgresChange` streams on a custom (non-`public`) schema, e.g. `schema: "library", table: "libros"` and `table: "libro_files"`.
4. Start two `for await` consumers on those streams in detached Tasks.
5. Call `try await channel.subscribeWithError()` (optionally wrapped in a retry-forever loop to avoid the SDK's "Maximum retry attempts reached" giving up after ~85 s).
6. Observe `client.realtimeV2.heartbeat` (AsyncStream<HeartbeatStatus>) and `channel.statusChange` from app code.

Expected: channel reaches `.subscribed` within seconds (as it does in v2.44.0).

Actual on v2.44.1 → v2.46.0: socket reaches `.connected` in ~2 s, but heartbeats time out every ~50 s (`sent` → `timeout`), triggering reconnects that kill in-flight JOINs. The channel stays in `.subscribing` until — by luck after multiple cycles — a JOIN slips through before its next heartbeat-timeout reconnect. Subscribe time observed between ~60 s and ~7 minutes depending on version.

### Code Sample

```swift
let channel = client.channel("library-realtime")
let libroStream = channel.postgresChange(AnyAction.self, schema: "library", table: "libros")
let fileStream  = channel.postgresChange(AnyAction.self, schema: "library", table: "libro_files")

// Consumers started before subscribe so they're ready when the channel attaches.
Task { for await c in libroStream { handle(c) } }
Task { for await c in fileStream  { handle(c) } }

// Retry-forever wrapper because subscribeWithError's internal max-retries
// (5 × ~10 s timeoutInterval) gives up before the first JOIN can land
// on broken versions.
Task {
    while !Task.isCancelled {
        do {
            try await channel.subscribeWithError()
            return
        } catch {
            try? await Task.sleep(for: .seconds(5))
        }
    }
}
```

### Relevant log output

```shell
T=0.0s   subscribe begin; channel = .subscribing; socket = .disconnected
T=2.1s   socket = .connected
T=26.6s  heartbeat sent          ← first heartbeat
T=51.1s  heartbeat timeout       ← second heartbeat fires, finds previous
                                   pendingHeartbeatRef set → timeout signalled
T=51.9s  socket = .connecting    ← SDK reconnects in response to heartbeat timeout
T=60.2s  socket = .connected
T=70.1s  subscribeWithError throws "Maximum retry attempts reached"
         (5 internal retries × ~10 s timeoutInterval exhausted inside the
         50 s window between socket-connect and heartbeat-timeout-reconnect)
... outer retry loop fires; pattern repeats ...
T=410.3s channel = .subscribed   ← eventually succeeds
T=410.3s+ heartbeats now ack instantly (`sent` → `ok`, same millisecond)
```

grdsdev:
## Summary

Fixes [SDK-959](https://linear.app/supabase/issue/SDK-959) / #999: the first `channel.subscribeWithError()` on a fresh socket stalls for 50s–7min (regression since v2.44.1 / #855). The socket connects but silently drops every incoming frame — heartbeat acks and `phx_join` replies never arrive — until a reconnect cycle happens to produce a healthy connection.

## Root Cause

All defects share one theme: **connection lifecycle events were not tied to a specific connection.**

### 1. Deaf socket from duplicate `handleConnected` (the stall itself)

`RealtimeClientV2.connect()` calls `handleConnected(conn:)` on every return — but `ConnectionManager.connect()` returns the *same* connection to every concurrent caller (`.connecting` / `.connected` / `.reconnecting` wait paths). So `handleConnected` runs multiple times for one physical socket whenever:

- **two channels subscribe concurrently at cold start** (both `ensureSocketConnected` → `connect()`), or
- **a subscribe retry calls `connect()` while an auto-reconnect is in flight** — the caller *and* the state-observer task each run `handleConnected` for the reconnected socket.

Each duplicate run of `handleConnected` called `listenForMessages`, which started a second consumer on `conn.events`. Because `AsyncStream` delivers each event to exactly one consumer, events were split between the two tasks — heartbeat acks and `phx_join` replies could be consumed by the task that was about to be cancelled, leaving the live task deaf.

**Fix:** `handleConnected` is now idempotent per connection (identity guard) — `listenForMessages` and `startHeartbeating` are called exactly once per socket.

### 2. `pendingHeartbeatRef` leaked across reconnects (the perpetual ~50s cycle)

Only the timeout path cleared `pendingHeartbeatRef`. After an error-driven reconnect with a heartbeat in flight, the new connection's first heartbeat misread the stale ref as a timeout and tore down the healthy socket — re-entering the cycle. **Fix:** cleared when adopting a new connection.

### 3. Stale-socket signals clobbered live state

`ConnectionManager.handleClose`/`handleError` accepted signals from any socket. A late `.close`/error surfacing from an old connection's message task (URLSession delivers these on a delegate queue, after the reconnect already established a new socket) flipped `.connected(newConn)` to `.disconnected`, orphaning a live connection. **Fix:** both now take the originating connection and ignore mismatches.

### 4. Channels went deaf after reconnect

`rejoinChannels()` called `subscribeWithError()` on channels that were already `.subscribed` — which is a no-op in `ChannelStateManager`. No new `phx_join` was sent on the new socket, so channels received nothing after every reconnect. **Fix:** `ChannelStateManager.resetForReconnect()` transitions `.subscribed`/`.subscribing`/`.unsubscribing` back to `.unsubscribed` (cancelling any in-flight task) before the rejoin sequence.

### 5. Server-initiated close did not trigger reconnect

`ConnectionManager.handleClose` only updated state without calling `initiateReconnect`. A server-driven WebSocket close (deploy restart, load-balancer timeout) left the SDK permanently disconnected. **Fix:** `handleClose` now calls `initiateReconnect` inside the `.connected` guard, matching the recovery path already used by `handleError`. SDK-initiated disconnects are unaffected because `disconnect()` transitions away from `.connected` before the close event arrives.

Application-level close codes (4000–4999) are excluded from auto-reconnect: a server close with code 4001 (invalid JWT) should not loop with the same bad token — the caller must re-authenticate first. Transport-level closures (1001 going-away, 1006 abnormal) still trigger reconnect.

### 6. `_clock` global race between parallel test classes

`ConnectionManager.initiateReconnect` and `RealtimeClientV2.startHeartbeating` read the module-level `_clock` global lazily at call time. When `RealtimeColdStartTests.setUp()` resets `_clock = ContinuousClock()` while a `RealtimeLifecycleTests` test runs in a parallel thread, heartbeat/reconnect sleeps could pick up the wrong clock and fire unexpectedly. **Fix:** both classes capture `_clock` as an instance property at init time; all internal sleeps use `self.clock`.

### 7. `WebSocket.events` must be a stored property (spydon review)

The protocol originally had a computed-property default for `events` in an extension — each read created a new `AsyncStream`. If `handleConnected` ran twice for the same socket, the second read started a second consumer, splitting the event stream (see #1). The extension default was also used by test doubles.

**Fix:** `onEvent` is removed from the protocol entirely. `events` is now a protocol requirement backed by a stored `let` property (and a matching `eventsContinuation`) created in `init` on both `URLSessionWebSocket` and `FakeWebSocket`. Reading `conn.events` any number of times returns the same stream.

## Testing

All regression tests verified to fail before their fix and pass after:

- `testRedundantConnect_doesNotDropIncomingFrames` — redundant `connect()` must not stop frame delivery (failed pre-fix: heartbeat `[.sent]`, never acked).
- **`RealtimeColdStartTests`** (new) — runs the reporter's exact flow (cold client, `postgresChange` streams, `subscribeWithError`) plus the two-channel concurrent variant against a transport with real connection latency and **asynchronous** frame delivery. Pre-fix: deterministic `maxRetryAttemptsReached` stall on iteration 1. Post-fix: subscribes in ~130ms across all iterations.
- `testPendingHeartbeatDoesNotLeakAcrossReconnect` — pre-fix: spurious `.timeout` + third connection after one reconnect.
- `testChannelsRejoinAfterReconnect` — pre-fix: channels stayed deaf after reconnect because `subscribe()` was a no-op on already-`.subscribed` state.
- `ConnectionManagerTests` — stale `handleClose`/`handleError` are ignored; current-connection signals still reconnect; application close codes (4001) do not reconnect.
- `RealtimeLifecycleTests` — foreground/background reconnect tests updated to match auto-reconnect-on-close behaviour.
- Full `RealtimeTests` suite: **210 tests, 0 failures**.

## Risk Assessment

- **Breaking changes**: None. Public API unchanged.
- **Behavior change**: duplicate `handleConnected` is now a no-op for setup; stale-socket signals no longer tear down healthy connections; server-initiated transport-level closes now trigger automatic reconnect (previously left permanently disconnected); application-level close codes (4000–4999) do NOT auto-reconnect (new).

## Linear

Closes SDK-959

### DIFF
--- a/Sources/Realtime/ChannelStateManager.swift
+++ b/Sources/Realtime/ChannelStateManager.swift
@@ -187,6 +187,37 @@ actor ChannelStateManager {
     }
   }
 
+  /// Reset channel state so it can rejoin after a reconnect.
+  ///
+  /// A channel that is already `.subscribed` when the socket reconnects
+  /// must re-send `phx_join` on the new socket. But ``subscribe()`` is a
+  /// no-op for `.subscribed` channels — so `rejoinChannels()` would silently
+  /// skip them without this reset.
+  ///
+  /// For `.subscribing`, the in-flight join task is waiting for a reply that
+  /// will never arrive on the dead socket; cancel it so a fresh `phx_join` is
+  /// sent on the new connection.
+  ///
+  /// For `.unsubscribing`, the wait loop in ``runUnsubscribe`` polls
+  /// `stateSubject.values`. Transitioning to `.unsubscribed` here wakes it up
+  /// immediately — the server's `phx_close` won't arrive on the dead socket
+  /// anyway.
+  func resetForReconnect() {
+    joinRef = nil
+    pushes = [:]
+    switch state {
+    case .subscribed:
+      updateState(.unsubscribed)
+    case .subscribing(let task):
+      task.cancel()
+      updateState(.unsubscribed)
+    case .unsubscribing:
+      updateState(.unsubscribed)
+    case .unsubscribed:
+      break
+    }
+  }
+
   // MARK: - Server signals
 
   /// Called when the server confirms the `phx_join` (system.ok or phx_reply
--- a/Sources/Realtime/ConnectionManager.swift
+++ b/Sources/Realtime/ConnectionManager.swift
@@ -102,22 +102,32 @@ actor ConnectionManager {
 
   /// Handle connection error and initiate reconnect.
   ///
-  /// - Parameter error: The error that caused the connection failure
-  func handleError(_ error: any Error) {
+  /// - Parameters:
+  ///   - error: The error that caused the connection failure
+  ///   - conn: The connection the error originated from. When provided and it
+  ///     isn't the current connection (e.g. a late error surfacing from a
+  ///     socket that has already been replaced by a reconnect), the error is
+  ///     ignored so it can't tear down a healthy connection.
+  func handleError(_ error: any Error, from conn: (any WebSocket)? = nil) {
     guard !(error is CancellationError) else {
       logger?.debug("CancellationError do not trigger reconnects.")
       return
     }
 
-    guard case .connected(let conn) = state else {
+    guard case .connected(let current) = state else {
       logger?.debug("Ignoring error in non-connected state: \(error)")
+      return
+    }
+
+    if let conn, conn !== current {
+      logger?.debug("Ignoring error from stale connection: \(error)")
       return
     }
 
     logger?.debug("Connection error, initiating reconnect: \(error.localizedDescription)")
 
     // Close the connection and update to disconnected before reconnecting
-    conn.close(code: nil, reason: "error: \(error.localizedDescription)")
+    current.close(code: nil, reason: "error: \(error.localizedDescription)")
     updateState(.disconnected)
 
     initiateReconnect(reason: "error: \(error.localizedDescription)")
@@ -130,11 +140,20 @@ actor ConnectionManager {
   /// - Parameters:
   ///   - code: WebSocket close code
   ///   - reason: WebSocket close reason
-  func handleClose(code: Int?, reason: String?) {
+  ///   - conn: The connection the close event originated from. When provided
+  ///     and it isn't the current connection (a `.close` frame from an old
+  ///     socket can arrive after a reconnect already established a new one),
+  ///     the event is ignored so it can't mark a healthy connection as
+  ///     disconnected.
+  func handleClose(code: Int?, reason: String?, from conn: (any WebSocket)? = nil) {
     let closeReason = "code: \(code?.description ?? "none"), reason: \(reason ?? "none")"
     logger?.debug("Connection closed by remote: \(closeReason)")
 
-    if case .connected = state {
+    if case .connected(let current) = state {
+      if let conn, conn !== current {
+        logger?.debug("Ignoring close from stale connection: \(closeReason)")
+        return
+      }
       updateState(.disconnected)
     }
   }
--- a/Sources/Realtime/ConnectionManager.swift
+++ b/Sources/Realtime/ConnectionManager.swift
@@ -17,6 +17,9 @@ actor ConnectionManager {
   private let headers: [String: String]
   private let reconnectDelay: TimeInterval
   private let logger: (any SupabaseLogger)?
+  // Captured at init time so parallel test classes that swap _clock cannot
+  // change the timing of an already-running client.
+  let clock: any _Clock
 
   /// Get current connection if connected, nil otherwise.
   var connection: (any WebSocket)? {
@@ -33,13 +36,15 @@ actor ConnectionManager {
     url: URL,
     headers: [String: String],
     reconnectDelay: TimeInterval,
-    logger: (any SupabaseLogger)?
+    logger: (any SupabaseLogger)?,
+    clock: any _Clock = _clock
   ) {
     self.transport = transport
     self.url = url
     self.headers = headers
     self.reconnectDelay = reconnectDelay
     self.logger = logger
+    self.clock = clock
   }
 
   @discardableResult
@@ -178,7 +183,7 @@ actor ConnectionManager {
 
   private func initiateReconnect(reason: String) {
     let reconnectTask = Task {
-      try await _clock.sleep(for: reconnectDelay)
+      try await clock.sleep(for: reconnectDelay)
       logger?.debug("Attempting to reconnect...")
       try await performConnection()
     }
--- a/Sources/Realtime/ConnectionManager.swift
+++ b/Sources/Realtime/ConnectionManager.swift
@@ -155,6 +155,7 @@ actor ConnectionManager {
         return
       }
       updateState(.disconnected)
+      initiateReconnect(reason: closeReason)
     }
   }
 
--- a/Sources/Realtime/ConnectionManager.swift
+++ b/Sources/Realtime/ConnectionManager.swift
@@ -160,6 +160,13 @@ actor ConnectionManager {
         return
       }
       updateState(.disconnected)
+      // Application-level close codes (4000–4999) typically signal auth or
+      // protocol errors. Reconnecting with the same bad token would loop;
+      // the caller must re-authenticate before reconnecting.
+      guard code.map({ $0 < 4000 }) ?? true else {
+        logger?.debug("Skipping reconnect for application close code \(code!)")
+        return
+      }
       initiateReconnect(reason: closeReason)
     }
   }
--- a/Sources/Realtime/RealtimeChannelV2.swift
+++ b/Sources/Realtime/RealtimeChannelV2.swift
@@ -157,6 +157,10 @@ public final class RealtimeChannelV2: Sendable, RealtimeChannelProtocol {
     await stateManager.unsubscribe()
   }
 
+  func resetForReconnect() async {
+    await stateManager.resetForReconnect()
+  }
+
   /// Build the `phx_join` payload from the current config and push it.
   /// Invoked by ``ChannelStateManager`` via the ``ChannelStateManager/JoinOperation``
   /// closure at the moment the state machine is ready to join.
--- a/Sources/Realtime/RealtimeClientV2.swift
+++ b/Sources/Realtime/RealtimeClientV2.swift
@@ -247,13 +247,30 @@ public final class RealtimeClientV2: Sendable, RealtimeClientProtocol {
   }
 
   private func handleConnected(conn: any WebSocket, isReconnect: Bool) {
-    mutableState.withValue { $0.connection = conn }
-    listenForMessages(conn: conn)
-    startHeartbeating()
+    // Set up the per-connection machinery exactly once. `handleConnected` can
+    // legitimately be invoked twice for the same physical connection — once
+    // synchronously by `connect()` and once by the state-observer task when a
+    // `connect()` call lands while an auto-reconnect is in flight. Reading
+    // `conn.events` a second time installs a fresh `onEvent` and, when the
+    // first event-stream's task is torn down, its `onTermination` nils that
+    // live `onEvent` out — leaving the socket connected but deaf (SDK-959).
+    // Guarding on connection identity keeps `listenForMessages` (and thus the
+    // `conn.events` read) to a single call per connection.
+    let isNewConnection = mutableState.withValue { state -> Bool in
+      if state.connection === conn { return false }
+      state.connection = conn
+      return true
+    }
+
+    if isNewConnection {
+      listenForMessages(conn: conn)
+      startHeartbeating()
+      flushSendBuffer()
+    }
+
     if isReconnect {
       Task { await rejoinChannels() }
     }
-    flushSendBuffer()
   }
 
   deinit {
--- a/Sources/Realtime/RealtimeClientV2.swift
+++ b/Sources/Realtime/RealtimeClientV2.swift
@@ -441,6 +441,7 @@ public final class RealtimeClientV2: Sendable, RealtimeClientProtocol {
 
   private func rejoinChannels() async {
     for channel in channels.values {
+      await channel.resetForReconnect()
       try? await channel.subscribeWithError()
     }
   }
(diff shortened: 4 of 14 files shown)


rpiacent:
**Field validation: this PR fixes #999 on real iOS devices.** A/B tested today on a production TestFlight app (Supabase realtime: private postgres_changes channels + private & public broadcast channels), iPhone + iPad, same network, same project.

**Protocol** (each round, repeated 3×): force-quit the app on BOTH devices → relaunch both → wait ~10s → toggle a state change on one device → check the other updates live.

| Build | SDK | Result |
|---|---|---|
| A | v2.46.0 (exact) | ❌ cross-device live updates dead after cold start, all cycles |
| B | this PR's head (`fb65ea0`) | ✅ instant sync after every cold start |

Extra scenario on build B: force-quit with an active connection on BOTH devices simultaneously → relaunch → everything resynchronized and live immediately.

Two additional data points that may help:
- **The bug appears iOS-conditional.** A macOS CLI harness (fresh `SupabaseClient` → `subscribeWithError()` on a public broadcast channel, 3 runs per version, release build) subscribes in **~0.8s on v2.44.0, v2.46.0 AND this PR's head** — we could not reproduce the stall from macOS at all, while it reproduces consistently on iOS devices.
- Build A also included an app-side change that guarantees a fresh auth token before subscribes, so stale-JWT effects are excluded as a confounder — the regression is the SDK version alone.

From our side this PR is merge-ready. Thanks for the fix — we're pinned to 2.44.0 until it ships in a release.

rpiacent:
**⚠️ Follow-up — important correction to our validation above: the fix DEGRADES over longer sessions.**

Extended field testing (same two iOS devices, same protocol):
- **This PR's head (`fb65ea0`)**: cold-start subscribes are fixed ✅, but **after some time running, cross-device live updates stop arriving** — reproducibly.
- **v2.44.0** running the same scenarios side-by-side: live updates keep working indefinitely. It remains the only version that is solid for us end-to-end.

So from our field data: this PR fixes the first-subscribe stall of #999 but appears to introduce (or leave) a longer-session regression — possibly around reconnect/heartbeat/token-refresh cycles rather than initial JOIN. Please hold off on treating our earlier comment as full validation; we're keeping our production pin on 2.44.0.

We'll gather more precise timings/conditions (time-to-degradation, foreground vs background, recovery behavior) and report back. Happy to test a revised branch on the same devices — turnaround on our side is fast.

rpiacent:
**Timing data you'll want (time-to-degradation):** on this PR's head, cross-device live updates stay healthy for roughly **2–3 minutes** after a fresh launch, then stop — reproducibly, on iOS devices, in foreground.

Two things that narrow it down:

1. **Not token-related.** 2–3 min is far short of the ~1h JWT TTL, so a stale/expired access token isn't the trigger.

2. **It's the socket/heartbeat lifecycle, not the initial JOIN.** With the SDK defaults (`heartbeatInterval` 25s, `reconnectDelay` 7s, `timeoutInterval` 10s, `disconnectOnSessionLoss` true), 2–3 min is ~5–7 heartbeat cycles — i.e. it degrades *after* a healthy run. Consistent with this PR fixing the first-JOIN stall while a separate defect remains in the reconnect / channel-rejoin path.

**Why it looks iOS-specific:** we never reproduced *any* of this on macOS — a release CLI harness subscribes in ~0.8s on 2.44.0, 2.46.0 AND this PR's head, and stays healthy indefinitely. That contrast points at something iOS hits that a stable-network macOS process doesn't — e.g. a network-path change (Wi-Fi assist / cellular handoff) triggering a socket reconnect whose channels then fail to rejoin (possibly the same surface as #579).

Happy to run a controlled foreground-only test on a fixed network to confirm the "channels don't rejoin after reconnect" hypothesis if useful. Our production pin stays on 2.44.0 (solid end-to-end for us) meanwhile — it's the only version where neither the stall nor the degradation appears.

grdsdev:
Fixed in ca9a3b6.

Root cause: `ChannelStateManager.subscribe()` is intentionally a no-op when the channel is already `.subscribed` — correct for normal operation (prevents duplicate joins), but it meant `rejoinChannels()` after reconnect silently skipped every channel that had already joined once.

The fix adds `resetForReconnect()` to `ChannelStateManager` that transitions `.subscribed` / `.subscribing` / `.unsubscribing` back to `.unsubscribed` before the rejoin sequence (cancelling any in-flight join task that was waiting on the dead socket). `rejoinChannels()` now calls this before `subscribeWithError()` so every channel re-sends `phx_join` on the new socket.

A regression test (`testChannelsRejoinAfterReconnect`) is included in the cold-start suite: it verifies that after a `phx_close` on socket 1, the channel reaches `.subscribed` on socket 2 and exactly one `phx_join` is sent on the new connection.

The 2–3 minute window you observed matches the default 25 s heartbeat interval × ~5–7 cycles — exactly how long it takes before a completely deaf channel becomes obvious in a live-sync scenario. The timing data is useful, appreciated.

rpiacent:
**Update — re-tested head `0fd759f` (incl. `ca9a3b6` + #1015), and I owe a correction to our earlier reports on this PR.**

Our previous "degrades after 2–3 min / never recovers" observations were **our own bug**, not the SDK. We had an app-side regression: a session/tenant re-activation tore down the previous realtime channel as a fire-and-forget task, racing the new same-topic subscribe and deafening it. That contaminated what we were attributing to the SDK.

With that fixed on our side (SDK held constant), we ran a clean A/B isolating the SDK on real iPhone + iPad:
- app fix + **supabase-swift 2.44.0** → passive cross-device live updates solid;
- app fix + **this PR head `0fd759f`** → **also solid**: the same passive cross-device live updates kept flowing, no 2–3-min degradation.

We also got one genuine cross-network data point on `0fd759f`: with one device on cellular and the other on Wi-Fi **at the same time**, cross-device live updates kept flowing. Note this is two devices on two networks, not a single-device Wi-Fi↔cellular handoff — so the on-device reconnect-on-remote-close path that #1015 targets isn't something I've exercised directly yet.

So from our side `0fd759f` looks good end-to-end on real iOS devices now. Apologies for the earlier noise — the degradation signal was our app bug, not this PR. We'll move off our 2.44.0 pin once this ships in a tagged release. Thanks for the fix.

spydon (commented):
Wouldn't it be better to move the deaf-socket fix down one layer to `WebSocket.events` so its teardown can't nil a handler it no longer owns?

spydon (commented):
After a further chat with the LLM:

> One small residual on the new fix
> 
> The extension default still contains the original footgun (onTermination { self.onEvent = nil }), kept "for test doubles." That's a reasonable tradeoff (a protocol extension can't hold the generation state), but it means a future production conformer that forgets to override events silently reintroduces SDK-959, with no compile-time signal. Optional hardening: drop the unsafe default entirely so every conformer is forced to implement the guarded version, or at minimum leave a // not safe for multi-read; production conformers must override warning on it. Not a blocker.
> 
> My other concern is still unaddressed ⚠️
> 
> ConnectionManager.handleClose still calls initiateReconnect on any remote close inside .connected, with no close-code discrimination:
> 
> if case .connected(let current) = state {
>   if let conn, conn !== current { ... return }
>   updateState(.disconnected)
>   initiateReconnect(reason: closeReason)   // fires for every close code
> }
> 
> This is the point from the last review that contradicts commit 29731d7's own stated "by design — auth expiry should not auto-reconnect" rationale. An auth-driven close (expired/invalid JWT) still produces: server closes → reconnect with the same bad token → server closes → loop, throttled only by reconnectDelay. The new commit is test-only, so nothing changed here. I'd still push for discriminating on close code (reconnect on 1006/1001/restart codes, but not on auth-related closes, or refresh the token before reconnecting) before this merges.

grdsdev:
> Wouldn't it be better to move the deaf-socket fix down one layer to WebSocket.events so its teardown can't nil a handler it no longer owns?

Actually we don't need the `onEvent` handler at all, we can default to only having the `events` AsyncStream as the only way of listening for incoming events, which fixes all those issues, did the change in https://github.com/supabase/supabase-swift/pull/1003/changes/35c1b3bcd85b807ddbe6e601fcd7a82c2982ee90

spydon (approved):
Lgtm!

PR #1010, opened 2026-06-10
feat(supabase): add W3C trace context propagation support

grdsdev:
## Summary

Adds protocol-based W3C trace context propagation to `SupabaseClient`, mirroring the `tracePropagation` option from supabase-js v2.106.0. A new `TraceContextProvider` protocol lets users inject `traceparent` / `tracestate` headers into every outgoing HTTP request without introducing any OTel library dependency in the SDK.

## Changes

- `Sources/Supabase/TraceContext.swift`: new `TraceContextProvider` protocol with DocC docs and an opentelemetry-swift usage example
- `Sources/Supabase/Types.swift`: `tracePropagation: (any TraceContextProvider)?` added to `GlobalOptions` (default `nil`, fully backward-compatible)
- `Sources/Supabase/SupabaseClient.swift`: trace headers injected in `adapt(request:)` (PostgREST / Storage / Functions), Auth fetch closure, and Realtime V2 default fetch closure — all four sub-clients covered
- `Tests/SupabaseTests/SupabaseClientTests.swift`: 3 new tests — header injection via PostgREST path, nil-provider no-op, Auth-path coverage; URLProtocol stub guarded with `NSLock` against the delegate-queue data race

## Usage

```swift
import OpenTelemetryApi

struct OTelTraceContextProvider: TraceContextProvider {
    func traceContext() -> [String: String] {
        guard let span = OpenTelemetry.instance.contextProvider.activeSpan else { return [:] }
        let traceId = span.context.traceId.hexString
        let spanId  = span.context.spanId.hexString
        return ["traceparent": "00-\(traceId)-\(spanId)-01"]
    }
}

let supabase = SupabaseClient(
    supabaseURL: url,
    supabaseKey: key,
    options: SupabaseClientOptions(
        global: .init(tracePropagation: OTelTraceContextProvider())
    )
)
// All sub-client requests now carry: traceparent: 00-<trace-id>-<span-id>-01
```

## Test plan

- [x] `testTracePropagationInjectsTraceParentHeader` — PostgREST path gets `traceparent`
- [x] `testTracePropagationIsNoOpWhenNil` — no header when provider is nil
- [x] `testTracePropagationInjectsHeaderIntoAuthRequests` — Auth path gets `traceparent`
- [x] Full `SupabaseTests` suite: 8/8 passing
- [x] Full package suite: all pre-existing tests unchanged (Storage/Realtime integration failures are pre-existing, require a live Supabase instance)

## Linear

Closes SDK-1043

### DIFF
--- a/Examples/Examples.xcodeproj/project.pbxproj
+++ b/Examples/Examples.xcodeproj/project.pbxproj
@@ -62,6 +62,8 @@
 		79FEFFC52B078D7900D36347 /* Models.swift in Sources */ = {isa = PBXBuildFile; fileRef = 79FEFFC42B078D7900D36347 /* Models.swift */; };
 		79FEFFC72B078FB000D36347 /* SwiftUIHelpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 79FEFFC62B078FB000D36347 /* SwiftUIHelpers.swift */; };
 		79FEFFC92B0797F600D36347 /* AvatarImage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 79FEFFC82B0797F600D36347 /* AvatarImage.swift */; };
+		AA110004AA110004AA110004 /* OpenTelemetryApi in Frameworks */ = {isa = PBXBuildFile; productRef = AA110002AA110002AA110002 /* OpenTelemetryApi */; };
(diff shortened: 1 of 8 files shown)


grdsdev:
Closing in favour of a redesign. The SDK is dropping support for Swift 5.10/6 and moving to Swift 6.1+ as the minimum, which will bring SPM package traits to the project. The trace propagation feature will be revisited as an optional trait-gated dependency once that migration lands, rather than the protocol-injection approach implemented here.

PR #1026, opened 2026-06-17
ci: wire up SDK compliance CI check

grdsdev:
Part of [SDK-992](https://linear.app/supabase/issue/SDK-992/ci-check-1-block-prs-adding-public-api-not-in-capability-matrix).

### `.github/workflows/validate-capabilities.yml` — updated to run on every PR

Expands the existing validate workflow to also run the public API symbol check on every PR by switching to the combined `validate-sdk-compliance.yml` reusable workflow from `supabase/sdk` and passing `language: swift`:

```yaml
jobs:
  validate:
    uses: supabase/sdk/.github/workflows/validate-sdk-compliance.yml@main
    with:
      language: swift
```

Previously this workflow only ran on PRs that touched `sdk-compliance.yaml` and only validated the file structure. Now on every PR it:

1. Validates the `sdk-compliance.yaml` structure (existing behaviour)
2. Checks out the PR branch and the base branch
3. Runs the Swift public API parser on both → computes new symbols
4. Checks each new symbol against `sdk-compliance.yaml`
5. Hard-fails if any new public symbol isn't registered

**Error output on failure:**
```
❌ Capability matrix check failed
New public API detected that is not in the capability matrix:
  - AuthClient.signInWithPassphrase (swift)

Register each symbol in sdk-compliance.yaml under the appropriate feature:
  auth.my_feature:
    status: implemented
    symbols:
      - AuthClient.signInWithPassphrase
```

### `sdk-compliance.yaml` — auth section annotated with symbols

Converts all implemented auth features from scalar form to object form with a `symbols` list mapping each capability matrix feature to its Swift symbol(s). For example:

```yaml
# before
auth.sign_in.sign_up: implemented

# after
auth.sign_in.sign_up:
  status: implemented
  symbols:
    - AuthClient.signUp
```

Covers all implemented auth groups: `sign_in` (12), `session` (9), `identities` (3), `mfa` (7), `admin` (6), `passkey` (2).

### `sdk-parse-ignore` — new file

Controls which files the public API parser visits. Without this file the parser would pick up test helpers, examples, and build artifacts as public symbols. Excluded paths:

```
.build/
DerivedData/
Examples/
docs/
Tests/
TestHelpers/
**/*Tests.swift
```

## How the check is progressive

The check only fails for **new** symbols (symbols added in this PR that didn't exist on the base branch). All existing public methods pass immediately with no annotation needed upfront.

## Test plan

- [x] `validate-compliance` passes on the updated YAML
- [x] E2E verified: PR with all symbols registered → passes; PR adding `AuthClient.signInWithMagicLink` (unregistered) → fails with correct message

### DIFF
--- a/.github/workflows/validate-capabilities.yml
+++ b/.github/workflows/validate-capabilities.yml
@@ -2,9 +2,6 @@ name: Validate SDK Compliance
 
 on:
   pull_request:
-    paths:
-      - 'sdk-compliance.yaml'
-      - '.github/workflows/validate-capabilities.yml'
 
 permissions:
   contents: read
@@ -12,3 +9,5 @@ permissions:
 jobs:
   validate:
     uses: supabase/sdk/.github/workflows/validate-sdk-compliance.yml@main
+    with:
+      language: swift
--- a/sdk-parse-ignore
+++ b/sdk-parse-ignore
@@ -1,0 +1,12 @@
+# Swift Package Manager artifacts
+.build/
+DerivedData/
+
+# Examples and documentation
+Examples/
+docs/
+
+# Tests
+Tests/
+TestHelpers/
+**/*Tests.swift
--- a/sdk-compliance.yaml
+++ b/sdk-compliance.yaml
@@ -7,48 +7,168 @@ features:
 
   # ── Auth ───────────────────────────────────────────────────────────────────
 
-  auth.sign_in.sign_up: implemented
-  auth.sign_in.sign_in_with_password: implemented
-  auth.sign_in.sign_in_with_oauth: implemented
-  auth.sign_in.sign_in_with_otp: implemented
-  auth.sign_in.sign_in_with_sso: implemented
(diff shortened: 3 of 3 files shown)


.github/workflows/validate-capabilities.yml
```diff
@@ -1,4 +1,4 @@
-name: SDK Compliance
+name: Validate SDK Compliance
 
 on:
```
QuintinWillison, line 3: I think you might also want this running on push to `main` (default branch).

sdk-parse-ignore
QuintinWillison, line 1: Not a biggie, merely a nit, but something makes me uneasy about this file's name not being dot(`.`)-prefixed (prior art: `.gitignore`) and also not having a file extension (akin to peer: `sdk-compliance.yaml`). Might just be me. Plus I recognise that this is part of your wider ecosystem so not easy to change, when I can see you're moving at pace.

QuintinWillison (approved):

PR #1027, opened 2026-06-18
test: E2E failure case — unregistered symbol

grdsdev:
This PR adds `AuthClient.signInWithMagicLink` which is NOT in `sdk-compliance.yaml`. The compliance check should hard-fail. **Delete after E2E testing.**


grdsdev:
E2E test complete — compliance check correctly blocks unregistered symbols. Closing test PR.

PR #1036, opened 2026-06-25
feat(postgrest): add type-safe PostgREST query layer via Swift Macros

grdsdev:
## Summary

Adds a type-safe PostgREST query layer to supabase-swift, allowing users to interact with their database using Swift types instead of raw strings — column names, table names, and query filters are all type-checked at compile time.

### What's new

**Swift Macros (`@Table`, `@SelectionOf`, `@Relationship`)** — Applied to a Swift struct, `@Table` synthesizes full `TableRepresentable` conformance including:
- `Insert` and `Update` nested types (with optional fields for `@Default`/nullable columns; `@PrimaryKey` fields excluded from inserts)
- `CodingKeys` with automatic `camelCase` → `snake_case` conversion (overridable via `@Column`)
- `columnName(for:)` using KeyPath identity for type-safe column lookup
- Read-only tables/views via `@Table(readOnly: true)`

`@SelectionOf(Table.self)` declares a partial column projection or join query. Fields annotated with `@Relationship(\Table.fkColumn)` generate PostgREST disambiguation syntax (`alias:tableName!fk_column(subselect)`) — the FK column is identified via a Swift KeyPath (type-checked at compile time) and the referenced type is inferred from the field's type annotation.

`@Table` enforces that `@Relationship` fields are never declared on it — joins belong exclusively on `@SelectionOf` structs.

**Typed query builders** — Four generic wrappers delegate to the existing string-based builders:
- `TypedPostgrestQueryBuilder<Table: TableRepresentable>` — insert, upsert, update, delete, select
- `TypedReadOnlyQueryBuilder<Table: ReadOnlyTableRepresentable>` — select-only (insert/update/delete are compile errors)
- `TypedPostgrestFilterBuilder<Table, Selection>` — KeyPath-based filters (`eq`, `neq`, `in`, `like`, `order`, `limit`, …)
- `TypedPostgrestTransformBuilder` / `TypedSingleResultBuilder` — transform and single-row execution

**Entry point** — `PostgrestClient.from(_ table: T.Type)` returns the appropriate typed builder based on whether `T` conforms to `TableRepresentable` or `ReadOnlyTableRepresentable`.

### Usage

```swift
// Define your table
@Table("messages")
struct Message {
  @PrimaryKey var id: UUID
  var senderId: UUID
  var body: String
}

// Declare a projection with a join
@SelectionOf(Message.self)
struct MessageWithSender {
  var id: UUID
  var body: String
  @Relationship(\Message.senderId) var sender: User
}
// MessageWithSender.selectString == "id,body,sender:users!sender_id(*)"

// Query with full type safety
let rows = try await supabase
  .from(Message.self)
  .select(MessageWithSender.self)
  .execute()
  .value  // [MessageWithSender]

// Filters use KeyPaths
let unread = try await supabase
  .from(Message.self)
  .select()
  .eq(\.senderId, value: userId)
  .execute()
  .value  // [Message]
```

### Architecture

The typed API lives entirely in `SupabaseSwiftMacros`, a separate opt-in library. `PostgREST` and `Supabase` have no dependency on it, so swift-syntax is only compiled for users who explicitly import `SupabaseSwiftMacros`.

```
SupabaseMacros (compiler plugin, internal)
    ↑
SupabaseSwiftMacros (opt-in library)
    ├── Protocols.swift         — SelectionRepresentable, ReadOnlyTableRepresentable, TableRepresentable
    ├── Macros.swift            — @Table, @SelectionOf, @PrimaryKey, @Default, @Column, @Relationship
    ├── TypedBuilders/          — TypedPostgrestQueryBuilder, TypedPostgrestFilterBuilder, …
    └── PostgrestClient+Typed.swift — from(_ table: T.Type) entry point

PostgREST  ←──── no swift-syntax dependency
Supabase   ←──── no swift-syntax dependency
```

### Diagnostics

Non-primitive `@SelectionOf` fields without `@Relationship`, `@Relationship` on a `@Table` field, and `let` bindings on `@Table` are all compile-time errors with descriptive messages.

### Testing

- 17 macro expansion tests (exact synthesized source assertions via `swift-macro-testing`)
- 8 Swift Testing tests verifying KeyPath → `snake_case` column name translation via inline-snapshot of actual HTTP request URLs

## Test plan

- [x] `swift test --filter SupabaseMacrosTests` — all 17 expansion/diagnostic tests pass
- [x] `swift test --filter PostgRESTTests` — all typed builder tests pass
- [x] `import PostgREST` alone (without `import SupabaseSwiftMacros`) does **not** pull in swift-syntax
- [ ] `@Table` struct with a `@Relationship` field is a compile error
- [ ] Non-primitive `@SelectionOf` field without `@Relationship` is a compile error
- [ ] `client.from(ReadOnlyView.self).insert(...)` is a compile error


grdsdev:
Closing this in favour of re-landing the implementation as **stage 1** of the PostgREST v3 RFC.

**The implementation is adopted, not rejected.** Reviewing this PR against a fresh design pass changed several of that design's decisions in this PR's favour:

- The separate `.macro` plugin plus opt-in library, with a test proving `import PostgREST` pulls in no swift-syntax — this replaces a weaker "the macro is optional" staging promise.
- `@Relationship(\Message.senderId)`. I had claimed relationships could not be KeyPaths because postgres-meta emits columns only. The foreign key *is* a column, so the KeyPath already exists in both directions. That retracted a whole invented requirement on the generator.
- `@PrimaryKey` / `@Default` synthesizing `Insert` and `Update`, replacing a default of `Insert = Self` that would have had callers inserting the primary key.
- `selectString` computed at expansion time, and `columnName<V>(for:)` instead of a `PartialKeyPath` dictionary.
- `@Column` as the input `CodingKeys` is generated from — I had dropped a per-column attribute on the mistaken grounds that `CodingKeys` already owned the name.
- Macro diagnostics and `swift-macro-testing` expansion tests as named requirements.
- The delivery order: typed layer first over today's builders, then the value-typed core swapped in beneath.

`ReadOnlyTableRepresentable` / `TableRepresentable` also arrived at the same read-only/writable protocol refinement the RFC reached independently.

**Why re-land rather than merge:**

1. It reports `CONFLICTING` and sits 31 commits behind `main`, so a rebase is required regardless.
2. Two structural changes touch every file here. The protocols need to move out of `PostgrestMacros` into `PostgREST`, because the query API is defined in terms of them and the stage-2 value-typed core cannot depend on the macro module. And they need renaming to `PostgrestSelection` / `PostgrestRelation` / `PostgrestWritableRelation`.
3. Merging as-is would publish `SelectionRepresentable` / `TableRepresentable` from the wrong module under names already rejected, then move and rename them. The package builds with `-enable-library-evolution`, so that is a binary break.

**Also salvaged** from the design spec on this branch, committed under `docs/superpowers/` before that path was gitignored and dropped from the tip: the user-facing command is `supabase gen types swift`; generated Postgres enums must conform to `PostgrestFilterValue`, or typed filters do not work on enum columns at all; and the Postgres-to-Swift type mapping the generator rewrite must agree with.

**References**

- RFC: https://linear.app/supabase/project/postgrest-v3-type-safe-query-api-rfc-ca49156abe11
- Spec, section 8 covers what was taken from this PR and section 8.1 the reasoning above: https://github.com/supabase/supabase-swift/blob/docs/postgrest-v3-design/docs/design/postgrest-v3.md

Stage 1 will lift these files with the protocols relocated and renamed.

PR #1048, opened 2026-07-02
fix(postgrest): remove invalid quotes from explain() plan media type

AndroidPoet:
When building the `Accept` header for `.explain()`, the plan format is wrapped in literal double quotes:

```
Accept: application/vnd.pgrst.plan+"text"; for=application/json; options=analyze|verbose;
```

Double quotes are not valid in the structured-syntax suffix of a media type, so PostgREST does not recognize `application/vnd.pgrst.plan+"text"` as the plan media type and the request does not behave as an EXPLAIN.

### Fix
Drop the quotes so the header matches the spec and the reference implementation. `postgrest-js` emits `application/vnd.pgrst.plan+${format}` (no quotes):

```
Accept: application/vnd.pgrst.plan+text; for=application/json; options=analyze|verbose;
```

### Notes
- The existing snapshot test `testExplain` had frozen the buggy `+"text"` output; it is updated in the same commit. Snapshot tests never hit a server, so it recorded the defect rather than validating it.
- All `PostgRESTTests` pass (93 tests, 0 failures).

### DIFF
--- a/Sources/PostgREST/PostgrestTransformBuilder.swift
+++ b/Sources/PostgREST/PostgrestTransformBuilder.swift
@@ -185,7 +185,7 @@ public class PostgrestTransformBuilder: PostgrestBuilder, @unchecked Sendable {
       .joined(separator: "|")
       let forMediaType = $0.request.headers[.accept] ?? "application/json"
       $0.request.headers[.accept] =
-        "application/vnd.pgrst.plan+\(format); for=\(forMediaType); options=\(options);"
+        "application/vnd.pgrst.plan+\(format); for=\"\(forMediaType)\"; options=\(options);"
     }
 
     return self
--- a/Sources/PostgREST/PostgrestTransformBuilder.swift
+++ b/Sources/PostgREST/PostgrestTransformBuilder.swift
@@ -185,7 +185,7 @@ public class PostgrestTransformBuilder: PostgrestBuilder, @unchecked Sendable {
       .joined(separator: "|")
       let forMediaType = $0.request.headers[.accept] ?? "application/json"
       $0.request.headers[.accept] =
-        "application/vnd.pgrst.plan+\"\(format)\"; for=\(forMediaType); options=\(options);"
+        "application/vnd.pgrst.plan+\(format); for=\(forMediaType); options=\(options);"
     }
 
     return self
--- a/Tests/PostgRESTTests/PostgrestTransformBuilderTests.swift
+++ b/Tests/PostgRESTTests/PostgrestTransformBuilderTests.swift
@@ -396,7 +396,7 @@ final class PostgrestTransformBuilderTests: PostgrestQueryTests {
     .snapshotRequest {
       #"""
       curl \
-      	--header "Accept: application/vnd.pgrst.plan+text; for=application/json; options=analyze|verbose;" \
+      	--header "Accept: application/vnd.pgrst.plan+text; for=\"application/json\"; options=analyze|verbose;" \
       	--header "Content-Type: application/json" \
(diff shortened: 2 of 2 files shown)


grdsdev:
Thanks @AndroidPoet for the fix, I cross checked the Swift implementation against JS and here are my findings, would be great to get all these fixed in this PR also.

## Code review findings

Reviewed against the [postgrest-js reference implementation](https://github.com/supabase/postgrest-js/blob/master/src/PostgrestTransformBuilder.ts).

### The core fix is correct ✓

Removing `\"` around `\(format)` is right — those were literal double-quote characters in the emitted header, not formatting. The JS reference uses bare `+${format}` with no quotes.

---

### One divergence from JS was missed

The JS reference **quotes the `for=` parameter value**, but the Swift fix leaves it unquoted:

```javascript
// postgrest-js
`application/vnd.pgrst.plan+${format}; for="${forMediatype}"; options=${options};`
//                                          ^^                ^^
```

```swift
// Swift after fix
"application/vnd.pgrst.plan+\(format); for=\(forMediaType); options=\(options);"
//                                         ^ not quoted
```

`application/json` contains `/`, which is not a valid [HTTP token character](https://httpwg.org/specs/rfc7231.html#media.type) under RFC 7231 §3.1.1.1 — parameter values containing `/` must be a `quoted-string`. PostgREST's parser is likely lenient here, but this is technically malformed and diverges from the reference.

Suggested fix:
```swift
"application/vnd.pgrst.plan+\(format); for=\"\(forMediaType)\"; options=\(options);"
```

---

### Other findings (pre-existing, not introduced by this PR)

**Double-call corruption** — calling `.explain()` twice nests the plan media type inside `for=`:
```swift
// explain().explain() produces:
// application/vnd.pgrst.plan+text; for=application/vnd.pgrst.plan+text; for=application/json; options=...
```
This is an edge case but worth a guard or note in the docs.

**`format` is untyped `String`** — the JS reference types this as `'json' | 'text'`. The current Swift signature accepts any string; passing `format: "xml"` silently produces an unrecognised media type. An enum would catch this at compile time.

**No test for `format: "json"`** — the single snapshot only covers the default `"text"` case. A regression in the `json` path would go undetected.

AndroidPoet:
Thanks for the thorough cross-check @grdsdev! Addressed the divergences:

**`for=` not quoted** — fixed in 2921a92. Now emits `for="application/json"`, matching the JS `for="${forMediatype}"`. Agreed it should be a quoted-string since `application/json` isn't a valid RFC 7231 token.

**No test for `format: "json"`** — added `testExplainWithJSONFormat` in the same commit, covering the `+json` plan path.

On the other two, I intentionally held off — let me know if you'd prefer them here anyway:

**Double-`explain()` nesting** — the JS reference does the same thing: it reads `this.headers.get('Accept')` and folds it into `for=`, so a second call nests there too. Since this PR is specifically about matching JS, I kept the behavior identical rather than diverging with a guard. Happy to add a doc note if you want to call it out.

**`format` as `String` vs `'json' | 'text'`** — an enum would be the right call for compile-time safety, but changing `format: String` to an enum is a source-breaking API change, and this is a stable 2.x release. Slipping it into a `fix:` (patch) PR felt wrong. I'd rather do it as a separate `feat!` PR (e.g. `ExplainFormat` enum) so it lands with the right semver bump — glad to open that if you agree.

grdsdev (approved):

grdsdev:
Thanks @AndroidPoet I agree the remaining should go a separate PR.

AndroidPoet:
Sounds good, thanks! I'll open the separate PR for the `ExplainFormat` enum.

This is my first contribution to supabase-swift and it's been a great experience — appreciate the detailed review.

PR #1050, opened 2026-07-02
feat(postgrest): add ExplainFormat enum for explain(format:)

AndroidPoet:
Follow-up to #1048, as discussed with @grdsdev.

The `explain()` `format` parameter is currently an untyped `String`, so a typo like `format: "jsom"` silently produces an unrecognised media type. The postgrest-js reference types this as `'json' | 'text'`.

### Change
Introduces a type-safe `ExplainFormat` enum for compile-time safety:

```swift
public enum ExplainFormat: String, Sendable {
  case text
  case json
}

builder.explain(analyze: true, format: .json)
```

**This is non-breaking.** The existing `String`-based overload is kept and marked `@available(*, deprecated)`, so current callers keep compiling (with a deprecation warning steering them to the enum). Both public overloads forward to a single private implementation, so the emitted header is unchanged.

### Notes
- Adds `testExplainWithJSONFormat` covering the `.json` path.
- All `PostgRESTTests` pass; `swift format lint --strict` clean.
- Stacked on top of #1048 — based off `main`, so the snapshots here reflect the current header formatting. Once #1048 merges I'll rebase and reconcile the json-format test.

### DIFF
--- a/Sources/PostgREST/PostgrestTransformBuilder.swift
+++ b/Sources/PostgREST/PostgrestTransformBuilder.swift
@@ -174,46 +174,6 @@ public class PostgrestTransformBuilder: PostgrestBuilder, @unchecked Sendable {
     wal: Bool = false,
     format: ExplainFormat = .text
   ) -> PostgrestTransformBuilder {
-    explain(
-      analyze: analyze,
-      verbose: verbose,
-      settings: settings,
-      buffers: buffers,
-      wal: wal,
-      formatValue: format.rawValue
-    )
-  }
-
-  /// Return `data` as the EXPLAIN plan for the query, using a string `format`.
-  @available(
-    *, deprecated, message: "Use the `ExplainFormat` overload, e.g. `explain(format: .json)`."
-  )
-  public func explain(
-    analyze: Bool = false,
-    verbose: Bool = false,
-    settings: Bool = false,
-    buffers: Bool = false,
-    wal: Bool = false,
-    format: String
-  ) -> PostgrestTransformBuilder {
-    explain(
-      analyze: analyze,
-      verbose: verbose,
-      settings: settings,
-      buffers: buffers,
-      wal: wal,
-      formatValue: format
-    )
-  }
-
-  private func explain(
-    analyze: Bool,
-    verbose: Bool,
-    settings: Bool,
-    buffers: Bool,
-    wal: Bool,
-    formatValue: String
-  ) -> PostgrestTransformBuilder {
     mutableState.withValue {
       let options = [
         analyze ? "analyze" : nil,
@@ -226,7 +186,7 @@ public class PostgrestTransformBuilder: PostgrestBuilder, @unchecked Sendable {
       .joined(separator: "|")
       let forMediaType = $0.request.headers[.accept] ?? "application/json"
       $0.request.headers[.accept] =
-        "application/vnd.pgrst.plan+\(formatValue); for=\"\(forMediaType)\"; options=\(options);"
+        "application/vnd.pgrst.plan+\(format.rawValue); for=\"\(forMediaType)\"; options=\(options);"
     }
 
     return self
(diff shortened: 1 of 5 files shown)


grdsdev:
Thanks for the PR! One design suggestion before this merges:

## `ExplainFormat` should be an open-ended struct, not a closed enum

A closed `enum` would force callers to fall back to the deprecated `String` overload the moment PostgREST adds a third plan format — exactly the API they migrated away from.

The suggested implementation:

```swift
public struct ExplainFormat: RawRepresentable, Hashable, Sendable {
  public let rawValue: String

  public init(rawValue: String) {
    self.rawValue = rawValue
  }

  /// Human-readable text output (default).
  public static let text = ExplainFormat(rawValue: "text")
  /// Machine-readable JSON output.
  public static let json = ExplainFormat(rawValue: "json")
}
```

Call sites stay identical (`.text`, `.json` still work via type inference), but a user on a future PostgREST version can pass `ExplainFormat(rawValue: "yaml")` without waiting for an SDK release.

It should also live at module top-level in `Sources/PostgREST/Types.swift` rather than nested inside `PostgrestTransformBuilder`, consistent with `CountOption`, `TextSearchType`, and the other PostgREST option types there.

---

Two other minor notes:
- The `testExplain` snapshot (line 399) will need updating on the rebase after #1048 — the PR description already calls this out, just flagging it here as well.
- The doc comment cross-reference ` ``explain(analyze:verbose:settings:buffers:wal:format:)`` ` is ambiguous between the two overloads; DocC will warn. Moving `ExplainFormat` to top-level (which removes the need for that cross-reference in its own doc comment) fixes this as a side effect.

AndroidPoet:
Great call, thanks! Done in the latest commit:

- **`ExplainFormat` is now an open-ended `RawRepresentable` struct** (`Hashable`, `Sendable`) instead of a closed enum, so a user on a newer PostgREST can pass `ExplainFormat(rawValue: "yaml")` without waiting for an SDK release. `.text` / `.json` call sites are unchanged.
- **Moved to module top-level in `Sources/PostgREST/Types.swift`**, next to `CountOption` / `TextSearchType`.
- That move also removed the enum's self-doc-comment that cross-referenced the ambiguous `explain(...)` overloads, so the DocC ambiguity warning is gone.
- The `testExplain` snapshot is already on the corrected `#1048` output since this branch has `main` merged in.

`swift format lint --strict` clean, all `PostgRESTTests` pass.

Sources/PostgREST/PostgrestTransformBuilder.swift
```diff
@@ -164,14 +164,55 @@ public class PostgrestTransformBuilder: PostgrestBuilder, @unchecked Sendable {
...
+  ) -> PostgrestTransformBuilder {
+    explain(
+      analyze: analyze,
+      verbose: verbose,
+      settings: settings,
+      buffers: buffers,
+      wal: wal,
+      formatValue: format
+    )
+  }
```
grdsdev: Can you move this deprecated message as an extension in the `Sources/PostgREST/Deprecated.swift` file?

We try to keep all deprecated files under that so it is easier to drop them on the next major release.

grdsdev (changes requested):

grdsdev:
## CI failure: capability matrix check

The **"Check public API against capability matrix"** job is failing because `ExplainFormat` is new public API that isn't registered in `sdk-compliance.yaml`.

### Fix

Apply this patch to `sdk-compliance.yaml`:

```diff
-  database.using_modifiers.explain: implemented
+  database.using_modifiers.explain:
+    status: implemented
+    symbols:
+      - ExplainFormat
```

AndroidPoet:
Done — moved the deprecated `explain(format: String)` overload into `Sources/PostgREST/Deprecated.swift` as an extension, matching the other deprecated declarations there so it's easy to drop at the next major. It forwards to the `ExplainFormat` overload via `ExplainFormat(rawValue:)`, and the main `explain` method is back to a single implementation.

grdsdev (approved):
LGTM, thanks.

PR #1051, opened 2026-07-02
ci: fix API stability check output and PR comment failure

grdsdev:
## What changed

- **`scripts/check-for-breaking-api-changes.sh`**: The script now captures the full `swift package diagnose-api-breaking-changes` output internally, pipes it to **stderr** (so it still appears in CI logs for debugging), and only writes the filtered `API breakage:` lines to **stdout**.
- **`.github/workflows/api-stability.yml`**:
  - Removed `2>&1` from the `API_OUTPUT` capture — only the filtered stdout (the actual breakage lines) is stored in the variable; the full build log streams naturally via stderr.
  - Removed the `echo "$API_OUTPUT"` that was re-printing the full log.
  - Replaced the unsafe `EOF` heredoc delimiter with `openssl rand -hex 8` to prevent `$GITHUB_OUTPUT` corruption if the output ever contains a bare `EOF` line.

## Why

The previous implementation captured the **entire build log** into `API_OUTPUT` — thousands of lines of package fetching, dependency resolution, and compilation warnings. This caused two failures:

1. **"Argument list too long"** — the `peter-evans/create-or-update-comment` action passed `api_output` as an environment variable to a Node.js process, which exceeded the OS argument length limit, making the step fail entirely.
2. **Unreadable output** — even when it didn't crash, the PR comment embedded the full build log, burying the ~15 actual `💔 API breakage:` lines in thousands of lines of noise.

## Reviewer notes

- The full build log is still visible in the CI step output (via stderr), so debugging information is not lost.
- The PR comment will now show only the compact list of breaking changes, e.g.:
  ```
  API breakage: var BucketOptions.fileSizeLimit has declared type change from Swift.String? to Storage.StorageByteCount?
  API breakage: var TransformOptions.resize has declared type change from Swift.String? to Storage.ResizeMode?
  ...
  ```

### DIFF
--- a/.github/workflows/api-stability.yml
+++ b/.github/workflows/api-stability.yml
@@ -33,13 +33,13 @@ jobs:
 
           echo "Checking for breaking API changes against $BASELINE_TREEISH"
 
-          # Run the breaking changes check and capture output
+          # Run the breaking changes check.
+          # stdout receives only the filtered "API breakage:" lines (see the script).
+          # stderr (full build log) flows directly to the CI log for debugging.
           set +e
-          API_OUTPUT=$(bash scripts/check-for-breaking-api-changes.sh 2>&1)
+          API_OUTPUT=$(bash scripts/check-for-breaking-api-changes.sh)
           API_EXIT_CODE=$?
           set -e
-
-          echo "$API_OUTPUT"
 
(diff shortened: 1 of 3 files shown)


spydon (approved):
Lgtm!

PR #1061, opened 2026-07-03
fix(postgrest): quote in() filter values containing reserved characters

AndroidPoet:
The `in()` filter joins values with commas without quoting, so any value containing a PostgREST reserved character (`,`, `(`, `)`) corrupts the filter:

```swift
.in("tags", values: ["a,b"])
// tags=in.(a,b) — matched as two values "a" and "b" instead of the single value "a,b"
```

### Change

Values containing `,`, `(`, or `)` are wrapped in double quotes before joining, matching the reference implementation in [postgrest-js](https://github.com/supabase/postgrest-js/blob/master/src/PostgrestFilterBuilder.ts) (`in()` quotes any element matching `[,()]`):

```swift
.in("tags", values: ["a,b", "c(d)", "plain"])
// tags=in.("a,b","c(d)",plain)
```

### Notes
- Adds `testInFilterQuotesValuesWithReservedCharacters` with a snapshot of the emitted request.
- postgrest-js also deduplicates values through a `Set` before joining; that is left out here to keep this fix focused (happy to add it in a follow-up if wanted).
- All `PostgRESTTests` pass; `swift format lint --strict` clean.

### DIFF
--- a/Sources/PostgREST/PostgrestFilterBuilder.swift
+++ b/Sources/PostgREST/PostgrestFilterBuilder.swift
@@ -1,4 +1,5 @@
 import Foundation
+import Helpers
 
 /// Filter builder for applying WHERE clauses to queries.
 ///
@@ -310,10 +311,7 @@ public class PostgrestFilterBuilder: PostgrestTransformBuilder, @unchecked Senda
     _ column: String,
     values: [any PostgrestFilterValue]
   ) -> PostgrestFilterBuilder {
-    let queryValues = values.map { value -> String in
-      let rawValue = value.rawValue
-      return rawValue.contains(where: { ",()".contains($0) }) ? "\"\(rawValue)\"" : rawValue
-    }
+    let queryValues = values.map { escapePostgRESTFilterValue($0.rawValue) }
     mutableState.withValue {
       $0.request.query.append(
         URLQueryItem(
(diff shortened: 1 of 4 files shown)


Sources/PostgREST/PostgrestFilterBuilder.swift
```diff
@@ -310,7 +310,10 @@ public class PostgrestFilterBuilder: PostgrestTransformBuilder, @unchecked Senda
     _ column: String,
     values: [any PostgrestFilterValue]
   ) -> PostgrestFilterBuilder {
-    let queryValues = values.map(\.rawValue)
+    let queryValues = values.map { value -> String in
+      let rawValue = value.rawValue
+      return rawValue.contains(where: { ",()".contains($0) }) ? "\"\(rawValue)\"" : rawValue
```
grdsdev: logic for escaping the in value was just merged to main added by this PR https://github.com/supabase/supabase-swift/pull/1063 which was for realtime `in` filer, let's reuse that code to escape the postgrest value also, move it to the `Helpers` module so both modules have access to it.

grdsdev (changes requested):

Sources/PostgREST/PostgrestFilterBuilder.swift
AndroidPoet: @grdsdev Done 👍 Moved the in() escaping from #1063 into Helpers as escapePostgRESTFilterValue(_:), and both PostgREST and Realtime now use it (Realtime's private copy removed). Merged main in first to pick up #1063. All tests pass. Let me know if you'd like a different name/location.

grdsdev (approved):

PR #1063, opened 2026-07-03
fix: support new realtime features

filipecabaco:

* support new system message to information replication connection is ready
* support new filters
* support select

https://linear.app/supabase/issue/SDK-1172/parityrealtime-add-postgres-changes-filter-builder-new-operators-and
https://linear.app/supabase/issue/SDK-1181/parityrealtime-add-replication-ready-broadcast-option-and-system-event

### DIFF
--- a/Sources/Realtime/RealtimeJoinConfig.swift
+++ b/Sources/Realtime/RealtimeJoinConfig.swift
@@ -56,21 +56,32 @@ public struct BroadcastJoinConfig: Codable, Hashable, Sendable {
   public var receiveOwnBroadcasts: Bool = false
   /// Configures broadcast replay from a specific timestamp
   public var replay: ReplayOption?
+  /// Instructs the server to emit a `system` event once the Postgres replication
+  /// connection backing this channel is established and ready to stream changes.
+  ///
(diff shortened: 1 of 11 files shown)


Sources/Realtime/RealtimeJoinConfig.swift
```diff
@@ -56,21 +56,53 @@ public struct BroadcastJoinConfig: Codable, Hashable, Sendable {
...
+
+  public func encode(to encoder: any Encoder) throws {
+    var container = encoder.container(keyedBy: CodingKeys.self)
+    try container.encode(acknowledgeBroadcasts, forKey: .acknowledgeBroadcasts)
+    try container.encode(receiveOwnBroadcasts, forKey: .receiveOwnBroadcasts)
+    try container.encodeIfPresent(replay, forKey: .replay)
+
+    if replicationReady {
+      try container.encode(replicationReady, forKey: .replicationReady)
+    }
```
grdsdev: If this behavior isn't really needed, there's no need to manually implement the decoder/encoder logic. Can't client send a `false` value for that?

grdsdev (approved):

PR #1064, opened 2026-07-03
build: add cSpell spell checker

grdsdev:
## Summary

- Adds [cSpell](https://cspell.org), pinned via a committed lockfile in `tools/node/package-lock.json` (installed with `npm ci`, the npm-native equivalent of a checksum-pinned Go tool), and wired into CI as its own `spell-check` job on `ubuntu-latest`.
- Fixes the genuine typos it surfaced across doc comments, log strings, and a few internal (non-public) identifiers: `stringfy` → `stringify`, `isLifecyleEvent` → `isLifecycleEvent`, plus assorted comment typos (`thorugh`, `recevied`, `Perfom`, `Lifecyle`, `Presense`, `Twillo`, `Emiter`, `Mutlipart`, etc.) and two British spellings (`recognise`, `stabilises`).
- Fixes a real prose/casing bug in the Realtime migration guide: `` `supabase.realtimev2.status` `` should have been `` `supabase.realtimeV2.status` ``.
- Records project-specific terms, proper nouns, and test fixture data (usernames, base64 fragments too short for the ignore pattern, non-ASCII i18n test strings, etc.) in `dictionary.txt` at the repo root.
- Adds `make spell-check` / `./scripts/spell-check.sh` for local use, documented in `AGENTS.md`.

Reference: [supabase/supabase-go#4](https://github.com/supabase/supabase-go/pull/4), which took the same approach (pinned cSpell via a committed npm lockfile) for the Go SDK.

## Test plan

- [x] `swift build` succeeds
- [x] `swift test` — same pass/fail counts as on unmodified `main` (pre-existing integration-test failures without a live Supabase instance are unrelated to this change)
- [x] `./scripts/spell-check.sh` passes with 0 issues across `**/*.swift` and `**/*.md`
- [x] Formatted only the touched Swift files with `swift-format`


spydon (commented):
Are none of these code typo fixes breaking changes?

grdsdev:
Good question — double-checked this:

- Ran `git diff origin/main...HEAD -- '*.swift' | grep -E "^[+-].*public "` — zero matches. No line adding/removing a `public` declaration exists in this diff.
- The two renamed functions are both non-public:
  - `stringfy` → `stringify` in [`PostgrestFilterValue.swift`](https://github.com/supabase/supabase-swift/blob/46b7e619/Sources/PostgREST/PostgrestFilterValue.swift#L79) and [`LoggerInterceptor.swift`](https://github.com/supabase/supabase-swift/blob/46b7e619/Sources/Helpers/HTTP/LoggerInterceptor.swift#L53) — plain (bare) `func`/`static func`, default `internal` access.
  - `isLifecyleEvent` → `isLifecycleEvent` in [`Defaults.swift`](https://github.com/supabase/supabase-swift/blob/46b7e619/Sources/Realtime/Deprecated/Defaults.swift#L102) — it lives inside `public enum ChannelEvent`, but the method itself has no `public` keyword, so it's internal even though the enum is public.
- The remaining renames (`testRpcWithGetMethodAndJSOBOject`→`...JSONObject`, `testgetUserById`→`testGetUserById`, `Cloudfare`→`Cloudflare`) are test-method names in the `Tests/` target, never shipped to consumers.
- Everything else was inside doc comments, log strings, or the markdown migration guide — prose, not identifiers.
- The **API stability check** job (`swift package diagnose-api-breaking-changes` against `main`) is passing on the latest commit as independent confirmation.

So no, none of these are breaking changes.

spydon (approved):

PR #1065, opened 2026-07-03
fix(realtime): send bare topic in broadcast REST body

AndroidPoet:
### Problem

`RealtimeChannelV2`'s REST broadcast path sends the channel's **full Phoenix topic** (`realtime:foo`) in the `/api/broadcast` request body, but the broadcast endpoint routes on the **bare sub-topic** (`foo`). As a result, REST-delivered broadcasts are routed to a topic namespace no WebSocket subscriber is listening on and are **silently dropped**.

This affects both REST entry points:
- `httpSend(event:message:)` — explicit REST delivery
- `broadcast(event:message:)` — the automatic REST fallback when the channel isn't subscribed

```swift
let channel = client.channel("room1")   // channel.topic == "realtime:room1"
try await channel.httpSend(event: "msg", message: ["x": 1])
// body → {"messages":[{"topic":"realtime:room1", ...}]}   ❌ misrouted
// expected                    {"topic":"room1", ...}       ✅
```

### Cause

`channel.topic` is set to `"realtime:\(topic)"`, and both REST payloads used it directly. This is a V2 regression — the deprecated V1 client strips the prefix (`subTopic`) for the REST body, and only the WebSocket frames use the full topic.

### Fix

Add an internal `subTopic` accessor that strips the leading `realtime:` prefix, and use it in the two REST body sites. WebSocket push frames continue to use the full `topic`.

### Tests

- Added `testHttpSend_bodyTopicHasNoRealtimePrefix` asserting the REST body carries the bare topic (verified to fail before the fix).
- Updated two existing tests (`testHttpSendSucceedsOn202Status`, `testBroadcastWithHTTP`) that asserted the prefixed topic.
- Full `RealtimeTests` suite passes (233 tests). No public API change.

### DIFF
--- a/Sources/Realtime/RealtimeChannelV2.swift
+++ b/Sources/Realtime/RealtimeChannelV2.swift
@@ -37,9 +37,7 @@ public final class RealtimeChannelV2: Sendable, RealtimeChannelProtocol {
 
   /// The channel's topic without the `realtime:` prefix, as expected by the
   /// broadcast REST endpoint (WebSocket frames use the full ``topic``).
-  var subTopic: String {
-    topic.hasPrefix("realtime:") ? String(topic.dropFirst("realtime:".count)) : topic
-  }
+  let subTopic: String
 
   @MainActor public private(set) var config: RealtimeChannelConfig
 
@@ -86,6 +84,8 @@ public final class RealtimeChannelV2: Sendable, RealtimeChannelProtocol {
     logger: (any SupabaseLogger)?
   ) {
     self.topic = topic
+    self.subTopic =
+      topic.hasPrefix("realtime:") ? String(topic.dropFirst("realtime:".count)) : topic
     self.config = config
     self.logger = logger
     self.socket = socket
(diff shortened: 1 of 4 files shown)


Sources/Realtime/RealtimeChannelV2.swift
```diff
@@ -35,6 +35,12 @@ protocol RealtimeChannelProtocol: AnyObject, Sendable {
 public final class RealtimeChannelV2: Sendable, RealtimeChannelProtocol {
   public let topic: String
 
+  /// The channel's topic without the `realtime:` prefix, as expected by the
+  /// broadcast REST endpoint (WebSocket frames use the full ``topic``).
+  var subTopic: String {
+    topic.hasPrefix("realtime:") ? String(topic.dropFirst("realtime:".count)) : topic
+  }
```
grdsdev: assign this once during channel initialization

grdsdev (changes requested):

grdsdev (approved):

grdsdev:
Thanks for the fix @AndroidPoet I'll wait for the release v2.49.0 https://github.com/supabase/supabase-swift/pull/1024 to go out before merging this one.

AndroidPoet:
@grdsdev Just wanted to thank you for letting me contribute. I hope I'm making a positive contribution to the project. 🙏

PR #1066, opened 2026-07-03
feat: drop Swift 5.10 support and require Swift 6.1+

grdsdev:
## Summary

- Bumps `swift-tools-version` to 6.1 and raises the minimum Xcode to 16.3. Per the [Support Policy](README.md#support-policy), dropping old Swift/Xcode versions is not a breaking change and ships in a minor release.
- The legacy pre-async/await Phoenix client under `Sources/Realtime/Deprecated/` predates Swift concurrency and isn't safe under Swift 6's strict-concurrency checking, and it can only be removed in a **major** release. To drop 5.10 without touching it:
  - The async/await implementation (`RealtimeClientV2`, `RealtimeChannelV2`, `PushV2`, the WebSocket transport, etc.) moves into a new `RealtimeV2` target, compiled under Swift 6's default language mode with real strict-concurrency enforcement.
  - The `Realtime` target keeps `Deprecated/` in place untouched, is pinned to `.swiftLanguageMode(.v5)` so it keeps compiling exactly as before, and depends on + `@_exported import`s `RealtimeV2` so `import Realtime` is unaffected for downstream consumers.
  - `RealtimeError` widened from `internal` to `package`, and the `ChannelEvent` enum moved from `Deprecated/Defaults.swift` into `RealtimeV2` — both are genuinely shared between old and new code and needed cross-target visibility.
  - All test targets are pinned to `.swiftLanguageMode(.v5)` too, since they were never in scope for strict-concurrency checking and the tools-version bump would have otherwise silently switched their default language mode to Swift 6, breaking unrelated pre-existing test code.
- CI: the Xcode 15.4 `macos-legacy` job is now pinned to the new minimum, Xcode 16.3 (on `macos-15`), and keeps Mac Catalyst coverage there instead of the main Xcode-latest (`macos`) job.
- `README.md`/`AGENTS.md` updated to reflect the new minimums.

## Test plan

- [x] `swift build` / `swift build --build-tests` succeed for the full package
- [x] `swift test` passes for `RealtimeTests`, `SupabaseTests`, `AuthTests`, `StorageTests`, `PostgRESTTests`, `FunctionsTests`, `HelpersTests`
- [x] Verified `Realtime` (v5 mode) does not affect `RealtimeV2`'s own Swift 6 language mode — introducing a Sendable violation in a scratch module reproduces a real compiler error regardless of the re-exporting module's language mode
- [x] `Deprecated/*.swift` is unchanged in behavior; only `ChannelEvent`'s physical location moved
- [ ] CI (this PR)


grdsdev:
This is a false positive. The listed symbols (`ChannelEvent`, `RealtimeChannelV2`, `RealtimeClientV2`, `RealtimeMessageV2`, the postgres/presence action types, etc.) didn't actually leave the public API — they moved from the `Realtime` target into a new `RealtimeV2` target, and `Realtime` re-exports `RealtimeV2` via `@_exported import`, so `import Realtime` still exposes all of them unchanged. The API-diff tool compares each target's own directly-declared symbols and doesn't account for `@_exported import` re-exports from a sibling target, so it sees them as "removed" from `Realtime` even though downstream consumers see no change.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

o-santi (approved):

PR #1070, opened 2026-07-03
fix(auth): decode callback URL params once, with form semantics

AndroidPoet:
### Problem

`extractParams(from:)` parses OAuth callback URL parameters, but decoded them incorrectly in two ways:

1. **Fragment values were double-decoded.** It read `components.fragment` (already percent-decoded by `URLComponents`) and split it, so a value like `100%2520off` became `100%20off` → `100 off`, and `+` was never converted to a space.
2. **The query path used `components.queryItems`**, which pre-decodes and does not convert `+` to a space — so PKCE errors delivered in the query (e.g. `?error_description=Email+link+is+invalid`) kept the literal `+`.

### Fix

Read the **raw** `percentEncodedFragment` / `percentEncodedQuery`, split on `&` / `=` first, then decode each component exactly once (`+` → space, then percent-decode). This matches auth-js `parseParametersFromURL` / `URLSearchParams` semantics for both the fragment and the query.

- `split(separator: "=", maxSplits: 1)` keeps values that contain `=`.
- A key with no `=` maps to an empty-string value (instead of being dropped).

### Tests

Added coverage for: `=`-in-value, percent-decoding, `+`-as-space (fragment and query), decode-after-splitting / no double-decode (fragment and query), and key-without-value. 5 of these fail on the current implementation; full `AuthTests` (200) pass.


grdsdev:
Reviewed this — the core fix (form-decoding once, via `percentEncodedFragment`/`percentEncodedQuery`) looks correct and the new test coverage for `extractParams` is solid. Found a couple of things worth a look before merge:

**1. Double-decode of `+` in `AuthClient.swift:862`/`:913` (blocker)**

Now that `extractParams`/`decodeFormComponent` already converts `+`→space and percent-decodes each value, these two call sites still re-apply `.replacingOccurrences(of: "+", with: " ")` on top of the already-decoded string:

```swift
// handleImplicitGrantFlow, AuthClient.swift:862
message: errorMessage.replacingOccurrences(of: "+", with: " ")

// handlePKCEFlow, AuthClient.swift:913
message: params["error_description"]?.replacingOccurrences(of: "+", with: " ")
```

This was a legitimate band-aid before this PR (the old query path never decoded `+`), but it's now a second pass that corrupts any `error_description` containing a literal `+` delivered as `%2B`. Example: fragment `#error_description=Rate%2Blimit%20exceeded` → `extractParams` correctly yields `"Rate+limit exceeded"` → these lines turn it into `"Rate limit exceeded"`, silently dropping the `+`. Since decoding is now centralized in `extractParams`, both of these `.replacingOccurrences` calls can just be deleted.

**2. Bare query key changes PKCE error control flow (edge case, worth a sanity check)**

Old `extractParams` dropped any pair that didn't split into exactly `key=value` (a bare key like `?error` with no `=` was silently discarded). New `extractParams` keeps bare keys with an empty-string value. For a URL like `myapp://callback?error&code=abc123`:
- Old: `params["error"]` is `nil` → `handlePKCEFlow`'s guard (`params["error"] != nil || ...`) is false → falls through and redeems the code normally.
- New: `params["error"]` is `""` (present) → guard trips → throws `AuthError.pkceGrantCodeExchange` instead of completing sign-in.

Low likelihood in practice since real IdPs send `error=<code>`, but it's a real control-flow change (not just a decoding change), and isn't covered by the new tests.

**3. Minor cleanup (non-blocking)**

The private `extractParams(from percentEncodedString:)` builds `[URLQueryItem]` only to immediately unpack it into a dictionary at the call site — the `URLQueryItem` wrapper adds an allocation with no payoff (its `Optional<String>` value is never used; empty string is always supplied instead). It's also named identically to the public `extractParams(from: URL)`, so `extractParams(from: fragment)` at the call site can read as a recursive call rather than a dispatch to a different overload. Building the `[String: String]` directly (or renaming the helper, e.g. `parsePairs(from:)`) would remove both the extra allocation and the ambiguity.

AndroidPoet:
Thanks for the thorough review all three addressed in f5583cc:

1. Double + decode (blocker) — removed both .replacingOccurrences(of: "+", with: " ") calls; decoding is now centralized in extractParams, so %2B survives as a literal +.
2. Bare-key control flow — the parser now drops both bare keys and empty-value pairs (guard parts.count == 2, !parts[1].isEmpty), matching the previous split semantics, so ?error&code=… no longer trips the error guard. Added coverage.
3. Cleanup — renamed the helper to parseFormEncodedPairs and build (name, value) pairs directly instead of allocating URLQueryItem.

grdsdev (approved):

grdsdev:
Hey @AndroidPoet  can you apply these changes to your branch to fix the spell checker job failure, https://github.com/supabase/supabase-swift/commit/62ee7e294167a5fc3df745ca45b801a2a2657edc

PR #1077, opened 2026-07-04
fix(realtime): keep current access token when token fetch fails in setAuth

AndroidPoet:
### Problem

`setAuth()` fetches the token via `try? await options.accessToken?()`, so if the callback **throws** (e.g. a transient failure while refreshing the JWT), the error is swallowed and `tokenToSend` becomes `nil`:

```swift
if tokenToSend == nil {
  tokenToSend = try? await options.accessToken?()
}
```

Because `setAuth()` is invoked on every heartbeat (`RealtimeClientV2.swift`, heartbeat path), a single flaky refresh then:

1. sets the client's `accessToken` to `nil`, and
2. pushes `access_token: null` to **every subscribed channel**

…silently de-authing all realtime subscriptions until the next successful refresh, and re-triggering on each heartbeat.

### Fix

Use `do/catch`: if the token fetch throws, log and return early — keep the current token and send no update to channels. This matches realtime-js, where a rejected `accessToken()` propagates and never results in a null push.

### Tests

- `testSetAuthKeepsCurrentTokenWhenAccessTokenFetchFails` — the client keeps its current token when the fetch throws.
- `testSetAuthDoesNotPushNullTokenToChannelsWhenFetchFails` — with a subscribed channel, a failing fetch pushes **no** `access_token` update over the socket.

Both fail on the previous implementation. Full `RealtimeTests` (235) pass.


grdsdev (approved):
LGTM! 

Just need to fix the merge conflict.

Sources/RealtimeV2/RealtimeClientV2.swift
```diff
@@ -735,12 +735,18 @@ public final class RealtimeClientV2: Sendable, RealtimeClientProtocol {
   /// is `nil`, the value is fetched from ``RealtimeClientOptions/accessToken`` if provided,
   /// or the token already stored on the client is used.
   ///
+  /// If the callback throws, the client keeps the current token and no update is sent to channels.
```
grdsdev:
```suggestion
  /// If ``RealtimeClientOptions/accessToken`` throws, the client keeps the current token and no update is sent to channels.
```

as a user facing doc comment, let's reference the callback so user knows "which callback throws".

grdsdev (changes requested):

grdsdev (approved):

PR #1080, opened 2026-07-06
fix(realtime): invalidate URLSession when web socket closes

AndroidPoet:
### Problem

`URLSessionWebSocket.connect` builds a delegate-backed `URLSession` for each connection, but that session was a local value that was never retained and never invalidated. A delegate-backed `URLSession` keeps a strong reference to its delegate until `invalidateAndCancel()` or `finishTasksAndInvalidate()` is called, so here the retain chain is:

```
URLSession -> _Delegate -> its callback closures -> the captured mutableState -> MutableState.webSocket (the socket)
```

`close()` and `_closeConnectionWithError()` only cancelled the task (`_task.cancel()`), never invalidating the session. As a result the session, its delegate, and the socket stayed alive after close. Since a fresh socket is created on every connect and reconnect, a long-lived client that reconnects on network changes leaks one session, delegate, and socket per reconnect.

### Fix

Retain the session on the socket and invalidate it once, in the single terminal close path, guarded so a duplicate close event cannot invalidate twice. `finishTasksAndInvalidate()` (rather than `invalidateAndCancel()`) lets any in-flight close frame flush, and it runs after the socket is logically closed so no needed delegate callback is dropped.

```swift
if case .close(let code, let reason) = event {
  let wasClosed = $0.isClosed
  $0.isClosed = true
  $0.closeCode = code
  $0.closeReason = reason
  if !wasClosed {
    session.finishTasksAndInvalidate()
  }
}
```

### Tests

Added `testSocketDeallocatesAfterClose` to `WebSocketTests` (gated on `canImport(Network)`). It stands up a small loopback WebSocket server, connects a real `URLSessionWebSocket`, holds a weak reference, closes it, drops the strong reference, and asserts the weak reference becomes nil. The test fails before this change (the socket stays alive because the session is never invalidated) and passes after. Full `RealtimeTests` suite passes (234 tests). No public API change.

### DIFF
--- a/Sources/RealtimeV2/WebSocket/URLSessionWebSocket.swift
+++ b/Sources/RealtimeV2/WebSocket/URLSessionWebSocket.swift
@@ -28,10 +28,12 @@ final class URLSessionWebSocket: WebSocket {
   ///   - _protocol: The negotiated WebSocket subprotocol, empty string if none.
   private init(
     _task: URLSessionWebSocketTask,
-    _protocol: String
+    _protocol: String,
+    session: URLSession
   ) {
     self._task = _task
     self._protocol = _protocol
+    self.session = session
 
     (events, eventsContinuation) = AsyncStream.makeStream()
 
@@ -104,7 +106,8 @@ final class URLSessionWebSocket: WebSocket {
       },
       onWebSocketTaskOpened: { session, task, `protocol` in
         mutableState.withValue {
-          $0.webSocket = URLSessionWebSocket(_task: task, _protocol: `protocol` ?? "")
+          $0.webSocket = URLSessionWebSocket(
+            _task: task, _protocol: `protocol` ?? "", session: session)
           $0.continuation.resume(returning: $0.webSocket!)
         }
       },
@@ -148,6 +151,7 @@ final class URLSessionWebSocket: WebSocket {
   let _task: URLSessionWebSocketTask
   /// The negotiated WebSocket subprotocol.
   let _protocol: String
+  private let session: URLSession
 
   /// Thread-safe mutable state for the WebSocket connection.
   struct MutableState {
@@ -299,9 +303,13 @@ final class URLSessionWebSocket: WebSocket {
 
       // Update state when connection closes
       if case .close(let code, let reason) = event {
+        let wasClosed = $0.isClosed
         $0.isClosed = true
         $0.closeCode = code
         $0.closeReason = reason
+        if !wasClosed {
+          session.finishTasksAndInvalidate()
+        }
       }
     }
   }
--- a/Sources/RealtimeV2/WebSocket/URLSessionWebSocket.swift
+++ b/Sources/RealtimeV2/WebSocket/URLSessionWebSocket.swift
@@ -298,7 +298,7 @@ final class URLSessionWebSocket: WebSocket {
   /// Triggers a WebSocket event and updates internal state if needed.
   /// - Parameter event: The event to trigger.
   private func _trigger(_ event: WebSocketEvent) {
-    mutableState.withValue {
+    let shouldInvalidate = mutableState.withValue {
       eventsContinuation.yield(event)
 
       // Update state when connection closes
@@ -307,10 +307,13 @@ final class URLSessionWebSocket: WebSocket {
         $0.isClosed = true
         $0.closeCode = code
         $0.closeReason = reason
-        if !wasClosed {
-          session.finishTasksAndInvalidate()
-        }
+        return !wasClosed
       }
+      return false
+    }
+
+    if shouldInvalidate {
+      session.finishTasksAndInvalidate()
     }
   }
 
--- a/Tests/RealtimeTests/WebSocketTests.swift
+++ b/Tests/RealtimeTests/WebSocketTests.swift
@@ -100,43 +100,51 @@ final class WebSocketTests: XCTestCase {
   // MARK: - URLSessionWebSocket Lifecycle Tests
 
   #if canImport(Network)
-    func testSocketDeallocatesAfterClose() async throws {
+    func testSocketsDeallocateAfterClose() async throws {
       let server = try LoopbackWebSocketServer()
       let port = try server.start()
       defer { server.stop() }
 
       let url = URL(string: "ws://127.0.0.1:\(port)")!
 
-      weak var weakSocket: URLSessionWebSocket?
+      var deallocations: [XCTestExpectation] = []
 
-      func connectCloseAndDrop() async throws {
+      for index in 0..<5 {
+        let deallocated = expectation(description: "socket \(index) deallocated")
+        deallocations.append(deallocated)
+
         let socket = try await URLSessionWebSocket.connect(to: url)
-        weakSocket = socket
+        objc_setAssociatedObject(
+          socket,
+          &deinitNotifierKey,
+          DeinitNotifier { deallocated.fulfill() },
+          .OBJC_ASSOCIATION_RETAIN_NONATOMIC
+        )
         socket.close(code: 1000, reason: nil)
       }
 
-      try await connectCloseAndDrop()
-
-      let deadline = Date().addingTimeInterval(5)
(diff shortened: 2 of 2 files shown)


Tests/RealtimeTests/WebSocketTests.swift
```diff
@@ -96,4 +96,95 @@ final class WebSocketTests: XCTestCase {
...
+          error: NSError(domain: "LoopbackWebSocketServer", code: -1)
+        )
+      }
+
+      return port.rawValue
+    }
+
+    private func receive(on connection: NWConnection) {
+      connection.receiveMessage { [weak self] _, _, _, error in
+        guard error == nil else { return }
```
grdsdev, line 204: The receive loop stops on any error but never sends a close-frame echo back. RFC 6455 §5.5.1 requires the responding endpoint to send its own close frame. Without it, `URLSessionWebSocketDelegate.urlSession(_:webSocketTask:didCloseWith:reason:)` is never called on the client — instead `didCompleteWithError` fires with a 1006 abnormal-closure code via the `onComplete` → `_connectionClosed(code:1006)` path. The test still passes, but it isn't exercising the clean 1000 close path it implies. On a future OS where `URLSession` waits for the full handshake before calling `didCompleteWithError`, the client would stall until the system socket timeout (~30–60 s).

Tests/RealtimeTests/WebSocketTests.swift
```diff
@@ -96,4 +96,95 @@ final class WebSocketTests: XCTestCase {
...
+
+      func connectCloseAndDrop() async throws {
+        let socket = try await URLSessionWebSocket.connect(to: url)
+        weakSocket = socket
+        socket.close(code: 1000, reason: nil)
+      }
+
+      try await connectCloseAndDrop()
+
+      let deadline = Date().addingTimeInterval(5)
```
grdsdev: The 5-second hard deadline can produce spurious CI failures if session invalidation takes longer on a loaded host, and can produce false passes if ARC frees the socket for an unrelated reason before the polling loop fires. A more deterministic design is a `XCTestExpectation` fulfilled from the socket's `deinit` (via a thin wrapper subclass or a captured closure that signals on deallocation), then `await fulfillment(of:timeout:)` — that way the test fails fast on failure and succeeds the instant the object is freed.

Tests/RealtimeTests/WebSocketTests.swift
```diff
@@ -96,4 +96,95 @@ final class WebSocketTests: XCTestCase {
...
+
+      weak var weakSocket: URLSessionWebSocket?
+
+      func connectCloseAndDrop() async throws {
+        let socket = try await URLSessionWebSocket.connect(to: url)
+        weakSocket = socket
+        socket.close(code: 1000, reason: nil)
+      }
+
+      try await connectCloseAndDrop()
```
grdsdev: The PR description says the leak manifests on every reconnect — each call to `connect()` leaks one session, delegate, and socket. A single connect-close cycle doesn't cover the accumulation pattern. Consider looping N times, collecting a weak ref per socket, and asserting all are nil after the loop. That directly exercises the described scenario and would have caught the original bug.

Tests/RealtimeTests/WebSocketTests.swift
```diff
@@ -96,4 +96,95 @@ final class WebSocketTests: XCTestCase {
...
+        self?.receive(on: connection)
+      }
+    }
+
+    func stop() {
+      queue.sync {
+        for connection in connections { connection.cancel() }
+        connections.removeAll()
+      }
+      listener.cancel()
```
grdsdev: Small race window: `newConnectionHandler` enqueues `self.connections.append(connection)` as an **async** task on `self.queue`, but `stop()` drains connections via `queue.sync` and then calls `listener.cancel()` **outside** the sync block. A connection accepted in the window between the sync block's return and `listener.cancel()` gets appended after `removeAll()` and is never cancelled — it keeps an `NWConnection` alive after `stop()` returns. Moving `listener.cancel()` inside the sync block (or stopping the listener first, then draining) closes the window.

Sources/RealtimeV2/WebSocket/URLSessionWebSocket.swift
```diff
@@ -299,9 +303,13 @@ final class URLSessionWebSocket: WebSocket {
 
       // Update state when connection closes
       if case .close(let code, let reason) = event {
+        let wasClosed = $0.isClosed
         $0.isClosed = true
         $0.closeCode = code
         $0.closeReason = reason
```
grdsdev, line 309: `session.finishTasksAndInvalidate()` is called while `mutableState`'s lock is held. `LockIsolated` uses `NSRecursiveLock` so same-thread re-entry won't deadlock today, but calling an external API under a lock makes future readers reason about lock ordering. The minimal fix is to capture a flag inside the lock and call `finishTasksAndInvalidate()` after releasing it:

```swift
let shouldInvalidate = mutableState.withValue { box -> Bool in
  let wasClosed = box.isClosed
  box.isClosed = true; box.closeCode = code; box.closeReason = reason
  return !wasClosed
}
if shouldInvalidate { session.finishTasksAndInvalidate() }
```

AndroidPoet:
Thanks for the thorough review. All five are addressed in 8ef322e:

**Session invalidation under the lock.** `_trigger` now returns a `shouldInvalidate` flag from inside the `mutableState` lock and calls `finishTasksAndInvalidate()` after the lock is released, so no external API runs while the lock is held.

**Fixed poll deadline / false pass.** Deallocation is now asserted from the socket's `deinit` through an `XCTestExpectation`, attached with a small `DeinitNotifier` via an associated object (the type is `final`, so a subclass was not an option). It fails fast on a real leak and passes the instant the objects are freed, with no arbitrary deadline.

**Single connect/close cycle.** The test loops over five connect/close cycles, one weak expectation per socket, and waits for all of them, so it exercises the per-reconnect accumulation the fix targets. It fails without the fix (all five time out) and passes with it.

**No close-frame echo.** The loopback server now detects the close opcode and sends its own close frame, so the client drives the clean 1000 close path rather than the 1006 abnormal-closure path.

**stop() race.** `stop()` now cancels the listener and drains connections inside the same `queue.sync`, and `newConnectionHandler` cancels any connection accepted after stop, closing the window.

Sources/RealtimeV2/WebSocket/URLSessionWebSocket.swift
AndroidPoet: Done. _trigger now returns a shouldInvalidate flag from inside the lock and calls finishTasksAndInvalidate() after the lock is released, so the session call no longer runs while the lock is held.

Tests/RealtimeTests/WebSocketTests.swift
AndroidPoet: Done in 8ef322e. The loopback server now detects the close opcode and sends its own close frame, so the client takes the clean 1000 close path rather than the 1006 abnormal-closure path.

Tests/RealtimeTests/WebSocketTests.swift
AndroidPoet: Done. Replaced the fixed deadline with a deinit-signalled XCTestExpectation, attached via a small DeinitNotifier associated object (the type is final, so a subclass was not an option). It fails fast on a real leak and passes the instant the socket is freed.

Tests/RealtimeTests/WebSocketTests.swift
AndroidPoet: Done. The test now loops five connect/close cycles with one expectation per socket and waits for all of them, exercising the per-reconnect accumulation. It times out on all five without the fix and passes with it.

Tests/RealtimeTests/WebSocketTests.swift
AndroidPoet: Done. stop() now cancels the listener and drains connections inside the same queue.sync, and newConnectionHandler cancels any connection accepted after stop, closing the window.

grdsdev (approved):

PR #1092, opened 2026-07-08
fix(postgrest): encode rpc GET/HEAD scalar params by JSON type

AndroidPoet:
### Problem

`rpc(_:params:head:get:count:)` builds the GET/HEAD query string from the `JSONSerialization` output with `String(describing:)`, which mis-encodes every non-string scalar:

```swift
for (key, value) in json {
  let formattedValue = (value as? [Any]).map(cleanFilterArray) ?? String(describing: value)
  url.appendQueryItems([URLQueryItem(name: key, value: formattedValue)])
}
```

- `Bool` becomes `1` / `0` (postgrest-js sends `true` / `false`)
- JSON `null` becomes `<null>`
- a nested object becomes the Objective-C `NSDictionary` debug string, e.g. `{\n    a = 1;\n}`

So `rpc("fn", params: ["enabled": true, "nested": ["a": 1]], get: true)` builds `?enabled=1&nested=%7B%0A%20%20%20%20a%20%3D%201;%0A%7D` instead of a usable query.

### Fix

Format each value by its actual JSON type:

- bool -> `true` / `false`
- null -> `null`
- number -> its numeric string
- array -> `{a,b,c}` (unchanged, via the existing `cleanFilterArray`)
- object -> JSON with sorted keys

Scalars and arrays now match postgrest-js. postgrest-js can't represent an object in a GET query at all (it stringifies to `[object Object]`); instead of sending garbage this encodes it as JSON, which PostgREST can cast for a `json`/`jsonb` argument. If you would rather object args fall back to POST (as supabase-js does), I am happy to change that.

### Tests

Added `testRpcWithGetMethodEncodesScalarParamsByJSONType`, which asserts a single `rpc(get: true)` call encodes `true`, `false`, `null`, an integer, and a nested object correctly. It fails before this change (bool -> `1`/`0`, null -> `<null>`, object -> the NSDictionary string) and passes after. Full `PostgRESTTests` suite passes (97 tests). No public API change.

### DIFF
--- a/Sources/PostgREST/PostgrestClient.swift
+++ b/Sources/PostgREST/PostgrestClient.swift
@@ -293,8 +293,7 @@ public final class PostgrestClient: Sendable {
       }
 
       for (key, value) in json {
-        let formattedValue = (value as? [Any]).map(cleanFilterArray) ?? String(describing: value)
-        url.appendQueryItems([URLQueryItem(name: key, value: formattedValue)])
+        url.appendQueryItems([URLQueryItem(name: key, value: queryValue(for: value))])
       }
 
     } else {
@@ -362,6 +361,33 @@ public final class PostgrestClient: Sendable {
     var configuration = configuration
     configuration.schema = schema
     return PostgrestClient(configuration: configuration)
+  }
+
+  private func queryValue(for value: Any) -> String {
+    switch value {
+    case let array as [Any]:
+      return cleanFilterArray(array)
+    case is NSNull:
+      return "null"
+    case let number as NSNumber:
+      if number === kCFBooleanTrue {
+        return "true"
+      }
+      if number === kCFBooleanFalse {
+        return "false"
+      }
+      return number.stringValue
+    case let string as String:
+      return string
+    default:
+      if let data = try? JSONSerialization.data(
+        withJSONObject: value, options: [.sortedKeys, .withoutEscapingSlashes]),
+        let json = String(data: data, encoding: .utf8)
+      {
+        return json
+      }
+      return String(describing: value)
+    }
   }
 
   private func cleanFilterArray(_ filter: [Any]) -> String {
--- a/Sources/PostgREST/PostgrestClient.swift
+++ b/Sources/PostgREST/PostgrestClient.swift
@@ -286,7 +286,7 @@ public final class PostgrestClient: Sendable {
     if head || get {
       method = head ? .head : .get
 
-      guard let json = try JSONSerialization.jsonObject(with: bodyData) as? [String: Any] else {
+      guard case .object(let json) = try AnyJSON.decoder.decode(AnyJSON.self, from: bodyData) else {
         throw PostgrestError(
           message: "Params should be a key-value type when using `GET` or `HEAD` options."
         )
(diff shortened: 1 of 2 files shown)


grdsdev:
## Review

Fix is correct for the reported case (bool/null/number/object encoding in `rpc(get:)`/`rpc(head:)` query params). Two things worth considering before merge, both pointing at the same root cause:

**1. `cleanFilterArray` still uses `String(describing:)` per element** ([PostgrestClient.swift:367](https://github.com/supabase/supabase-swift/blob/fix/postgrest-rpc-get-scalar-encoding/Sources/PostgREST/PostgrestClient.swift#L367))

The exact bug this PR fixes for top-level scalars is still present for array elements:

```swift
rpc("fn", params: ["flags": [true, false, AnyJSON.null]], get: true)
// still emits {1,0,<null>} instead of {true,false,null}
```

**2. `queryValue(for:)` re-derives JSON typing dynamically instead of using the codebase's own `AnyJSON` enum** ([PostgrestClient.swift:363](https://github.com/supabase/supabase-swift/blob/fix/postgrest-rpc-get-scalar-encoding/Sources/PostgREST/PostgrestClient.swift#L363))

`Sources/Helpers/AnyJSON/AnyJSON.swift` already models exactly these JSON cases (`.null`, `.bool`, `.integer`, `.double`, `.string`, `.object`, `.array`), Codable and exhaustive. The current fix instead switches on `Any`/`NSNumber`, including a `number === kCFBooleanTrue` pointer-identity check to distinguish `Bool` from `Int`. That's a well-known fragile Objective-C bridging trick — its behavior isn't guaranteed identical between Darwin and swift-corelibs-foundation, and this package builds on Linux (per AGENTS.md).

Suggest decoding `bodyData` into `[String: AnyJSON]` (instead of `JSONSerialization.jsonObject(...) as? [String: Any]`) and switching over the enum. That:
- removes the CF-bridging dependency entirely,
- fixes both call sites (top-level values and array elements) with one recursive code path, since `.array` already holds typed `AnyJSON` elements,
- matches how the rest of the module avoids raw `Any` (e.g. `PostgrestFilterValue`).

Net: smaller, more idiomatic, and more future-proof than extending the `Any`-based switch further.

AndroidPoet:
Good call on both, done in d123539.

I decode `bodyData` into `[String: AnyJSON]` and switch over the enum, so there is one recursive `queryValue(for:)` path now. That fixes the array-element case you flagged: `rpc("fn", params: ["flags": [true, false, .null]], get: true)` now emits `{true,false,null}` instead of `{1,0,<null>}`. `cleanFilterArray` and the `NSNumber`/`kCFBoolean` bridging are gone.

Objects still get JSON-encoded with sorted keys via a local `JSONEncoder` (`[.sortedKeys, .withoutEscapingSlashes]`) rather than `AnyJSON.encoder`, since the shared encoder only applies `.sortedKeys` under testing and would escape slashes in release, so the query string would be non-deterministic otherwise.

Extended `testRpcWithGetMethodEncodesScalarParamsByJSONType` to cover the array-element case too.

grdsdev (approved):

PR #1095, opened 2026-07-08
test(postgrest): migrate PostgRESTTests to Swift Testing, keep Mocker (Phase 4)

grdsdev:
## Summary

Migrates all 10 files in `Tests/PostgRESTTests` (~2900 lines) from XCTest to Swift Testing (`#expect`/`#require`/`@Test`/`@Suite`), per the Phase 0 decision (SDK-1247). Keeps Mocker for HTTP mocking — Replay isn't proven yet for all our cases, so we're sticking with Mocker for now (may revisit later). This is Phase 4 of the migration tracked in SDK-435.

## Changes

- All 10 test files converted to Swift Testing; zero `import XCTest` remains in the target.
- Every `.snapshotRequest { curl... }` curl-body assertion is kept as-is — Mocker gives full query-string/body/header verification for free, so none of the extra matcher machinery an equivalent Replay port would need is necessary here.
- `PostgresQueryTests.swift`, an `XCTestCase` base class shared via inheritance by 5 other test suites, is replaced by a composed `PostgrestQueryFixture` struct (renamed to `PostgrestQueryFixture.swift`), since Swift Testing suites don't support shared state through subclassing.
- `PostgrestBuilderTests` is a plain `struct` (no `deinit`/class needed): the process-global `_clock` test seam it used to reset was already removed by a separate refactor that injects the clock explicitly per test via `makeSUTWithCustomFetch`.
- `BuildURLRequestTests` keeps its `assertSnapshot(as: .curl)` request-shape snapshots, switching `#file` → `#filePath` since Swift Testing resolves `#file` to a module-relative path (breaks snapshot directory resolution under both `swift test` and `xcodebuild`).
- Carries forward test coverage added to `main` after this branch's original base: `notIn`/`dryRun`/`maybeSingle` modifier tests and the `select(count:)` Prefer-header fix test.
- `PostgRESTTests` moves off the blanket `.swiftLanguageMode(.v5)` pin onto full Swift 6 language mode, matching production targets. Drops the unused `InlineSnapshotTesting` dependency. No new package dependency.

## Cross-target Mocker race (found while verifying this PR)

The 5 suites built on `PostgrestQueryFixture` share Mocker's mock registry, so — same as `StorageBucketAPITests`/`StorageFileAPITests` — they're nested under a `PostgrestMockerTests(.serialized)` namespace so they don't run concurrently with each other.

That alone wasn't enough: Mocker's registry is process-global across the *entire test binary*, not per target. Running `PostgRESTTests` and the already-merged `StorageTests` together reliably failed (`swift test --filter "StorageTests|PostgRESTTests"`), even though each passed cleanly alone — Swift Testing schedules suites from every compiled test target against one shared executor, so `StorageMockerTests` and `PostgrestMockerTests` could still interleave and race on Mocker's shared registry.

Added `Sources/TestHelpers/MockerSerialization.swift`: a small process-wide async lock exposed as a `.mockerSerialized` Swift Testing trait, applied directly to every concrete Mocker-backed suite in both `StorageTests` and `PostgRESTTests`. It has to be applied suite-by-suite (`isRecursive = false`), not recursively via the enclosing namespace enums — marking it recursive there crashes Swift Testing's trait-application graph (stack overflow) when the namespace's only nested suites are declared via cross-file `extension`.

This touches `StorageTests` (already merged, outside this PR's nominal scope) because leaving it out would mean merging this PR immediately reintroduces full-suite flakiness on `main`.

**Note:** the pending Functions PR (#1094), once merged, will need the same `.mockerSerialized` treatment on `FunctionsClientTests` — it has its own Mocker-backed Swift Testing suite that will race against these two the same way.

## Test plan

- [x] `swift test --filter PostgRESTTests`: 105/105 tests, 10 suites, run 6+ times consecutively — no flakiness (up from the original 97 due to tests added to `main` since).
- [x] `swift test --filter "StorageTests|PostgRESTTests"`: run 5+ times — reliably failed before the cross-target fix, clean after.
- [x] Full `swift test --skip IntegrationTests`: run 3+ times, no regressions.
- [x] `./scripts/spell-check.sh` and `./scripts/format.sh`: clean.

## Linear

Closes SDK-1251


spydon (approved):

grdsdev:
Addressed in 334b571c:
- Added `Issue.record("Expected error to be thrown")` after `execute()` in the 4 flagged error-path tests (`executeWithNonSuccessStatusCode`, `executeWithNonJSONError`, `rpcWithGetMethodAndNonJSONObjectShouldThrowError`, `rpcWithHeadMethodAndNonJSONObjectShouldThrowError`) — all previously passed vacuously if `execute()` stopped throwing.
- Renamed `cSV`/`cSVWithStripNullsThrowsError` → `csv`/`csvWithStripNullsThrowsError`.
- Checked Copilot's "unused `import HTTPTypes`" claim on `PostgrestBuilderTests.swift` — it's a false positive, removing it breaks the build (`.init("apikey")!` implicitly constructs an `HTTPField.Name`, which needs the module in scope even without an explicit type name at the call site). Left it in place.

Verified: 105/105 tests pass (3 runs), full suite green, format/spell-check clean.

PR #1097, opened 2026-07-08
test(integration): migrate IntegrationTests to Swift Testing (Phase 6)

grdsdev:
## Summary

Migrates all 10 files in `Tests/IntegrationTests` from XCTest to Swift Testing (`@Suite`/`@Test`), per the SDK-435 migration plan and the Phase 0 conventions decided in SDK-1247. This target has the heaviest `setUp`/`tearDown` usage in the repo (11 total) and orchestrates a real local Supabase instance rather than mocking, so it needed its own lifecycle patterns rather than reusing Phase 2-4's Mocker→Replay work (unrelated — no Mocker/Replay usage here at all).

## Changes

- `XCTSkipUnless(INTEGRATION_TESTS)` in every `setUp` becomes a `@Suite(.enabled(if:))` trait, evaluated once per suite instead of once per test method.
- `setUp`/`tearDown` become `init()`/`deinit()`. Files with only read-only shared state per test (the 4 Postgrest files, `PostgrestIntegrationTests`, `AuthClientIntegrationTests`, `StorageClientIntegrationTests`) convert to `struct`s — no teardown needed.
- `StorageFileIntegrationTests` and `RealtimeIntegrationTests` need real async teardown (delete a test bucket / disconnect realtime channels), so they stay `final class` with a `deinit` that fires cleanup in a best-effort detached `Task`, mirroring the original `try?`-swallowed `tearDown` semantics (documented inline — this is an accepted tradeoff for ephemeral local test resources, not correctness-critical).
- `RealtimeIntegrationTests` additionally needs `@Suite(.serialized)` because it swaps the process-global `_clock` seam (`Sources/Helpers/_Clock.swift`) per test — mirrors the Phase 4 `PostgrestBuilderTests` precedent — plus `Sendable` conformance on the class since `withMainSerialExecutor`'s closure parameter is `@Sendable` outside DEBUG/Concurrency-availability builds under Swift 6 strict checking.
- `AuthClientIntegrationTests`' custom `XCTAssertAuthChangeEvents` helper (built on `XCTestExpectation`) is replaced with a `LockIsolated` + short polling loop, since Swift Testing has no direct expectation-fulfillment equivalent.
- Moves `IntegrationTests` out of the blanket Swift 5 language-mode pin (`Package.swift`) into full Swift 6 mode, matching production targets' `swiftSettings`. No other `Sendable`/isolation diagnostics needed fixing beyond the `RealtimeIntegrationTests` case above.

## Test plan

- [x] `swift build --target IntegrationTests` compiles clean under full Swift 6 mode.
- [x] `swift test --filter IntegrationTests`: all 8 `INTEGRATION_TESTS`-gated suites (Postgrest x4, `PostgrestIntegrationTests`, `AuthClientIntegrationTests`, `StorageClientIntegrationTests`, `StorageFileIntegrationTests`) correctly skip via the new `.enabled(if:)` trait — confirms that gate is equivalent to the old per-test `XCTSkipUnless`.
- [ ] `RealtimeIntegrationTests` fails in this environment with `CancellationError` (no local Supabase instance reachable) — this is pre-existing: its `XCTSkipUnless` was already commented out in the original XCTest file (predates this PR), so it has always attempted a live connection unconditionally. Not a regression from this migration. Full live verification (`./scripts/test-integration.sh` against `supabase start`) needs a local Supabase instance and wasn't run in this environment.

## Linear

Closes SDK-1253

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,7 +177,7 @@ jobs:
       - name: "Run Integration Tests"
         env:
           INTEGRATION_TESTS: 1
-        run: swift test --filter IntegrationTests
+        run: swift test --filter IntegrationTests --no-parallel
       - name: "Stop Supabase"
         working-directory: Tests/IntegrationTests
         if: always()
--- a/Package.swift
+++ b/Package.swift
@@ -243,6 +243,7 @@ let package = Package(
 // phase lands (see SDK-435).
 let swift6TestTargets: Set<String> = [
   "SupabaseTests", "HelpersTests", "StorageTests", "PostgRESTTests", "AuthTests", "FunctionsTests",
+  "IntegrationTests",
 ]
 
 for target in package.targets {
--- a/Tests/IntegrationTests/AuthClientIntegrationTests.swift
+++ b/Tests/IntegrationTests/AuthClientIntegrationTests.swift
@@ -7,9 +7,10 @@
 
 import ConcurrencyExtras
 import CustomDump
+import Foundation
 import InlineSnapshotTesting
 import TestHelpers
-import XCTest
+import Testing
 
 @_spi(Experimental) @testable import Auth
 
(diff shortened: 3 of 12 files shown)


spydon (approved):

grdsdev:
@copilot resolva os conflitos de merge nesta branch.

PR #1110, opened 2026-07-10
fix(realtime): rejoin channels concurrently on reconnect

AndroidPoet:
### Problem

On reconnect, `rejoinChannels()` re-subscribes each channel one at a time, awaiting each channel's join to be confirmed by the server before starting the next:

```swift
for channel in channels.values {
  await channel.resetForReconnect()
  try? await channel.subscribeWithError()
}
```

`subscribeWithError()` suspends until the server confirms the join (or the attempt times out and retries). So if one channel is slow to rejoin, every channel after it in the loop waits, and none of them send `phx_join` until the slow one resolves. With several channels, one stuck join stalls the rest for up to the full join timeout.

realtime-js rejoins all channels at once, so a single slow channel cannot block the others.

### Fix

Rejoin the channels concurrently with a task group, so each channel resets and re-subscribes independently.

### Tests

Added `testRejoinChannelsSubscribesConcurrently`: subscribes two channels, forces a reconnect where the server does not confirm the joins, and asserts both channels send `phx_join` on the new connection. Before this change only one channel's join is sent (the other is blocked behind it); after, both are sent. Full `RealtimeTests` suite passes (241 tests). No public API change.


Sources/RealtimeV2/RealtimeClientV2.swift
```diff
@@ -536,9 +536,13 @@ public final class RealtimeClientV2: Sendable, RealtimeClientProtocol {
   }
 
   private func rejoinChannels() async {
-    for channel in channels.values {
-      await channel.resetForReconnect()
-      try? await channel.subscribeWithError()
+    await withTaskGroup(of: Void.self) { group in
```
grdsdev: Lets use DiscardingTaskGroup since a result isn't needed.

```suggestion
    await withDiscardingTaskGroup(of: Void.self) { group in
```

grdsdev (changes requested):

grdsdev (approved):

grdsdev:
Ok, so it seems `withDiscardingTaskGroup` isn't available in the current support OS versions, please can you revert that? Sorry.

AndroidPoet:
Merged `main` in to clear the conflict, and ported the regression test to Swift Testing since `AGENTS.md` now requires it (it also picks up the `waitUntil` helper from `Tests/RealtimeTests/TestSupport.swift` instead of an inline polling loop).

One note on the `withDiscardingTaskGroup` suggestion: I had to revert it. It is available from iOS 17, macOS 14, watchOS 10 and tvOS 17, while this package targets iOS 16, macOS 13, watchOS 9 and tvOS 16, so it fails to build:

```
error: 'withDiscardingTaskGroup(returning:isolation:body:)' is only available in macOS 14.0 or newer
error: 'addTask(priority:operation:)' is only available in macOS 14.0 or newer
```

The applied suggestion also used `of:`, which that API does not take. I have kept `withTaskGroup(of: Void.self)`, which matches `removeAllChannels()` just above. Happy to gate it behind `#available` if you would rather use the discarding group where it is available.

263 RealtimeTests green; the test still fails without the fix (only `realtime:topic-a` rejoins, `topic-b` is blocked behind it).

PR #1117, opened 2026-07-12
feat(realtime): add serverTrustHandler to RealtimeClientOptions for WebSocket certificate pinning

6Ric6:
## Motivation

The Realtime WebSocket transport (`URLSessionWebSocket`) creates its own internal `URLSession`, which bypasses any `URLSessionDelegate`-based certificate pinning an app configures on its main Supabase client session. Apps that SPKI-pin their REST / Auth / Storage traffic therefore have **no way to pin the Realtime WebSocket connection** — the channel silently connects unpinned, a security downgrade for apps where pinning is a requirement. There's currently no public hook to evaluate server trust on the Realtime session.

## What this does

Adds an optional `serverTrustHandler` closure to `RealtimeClientOptions`, forwarded to the WebSocket `URLSession`'s delegate and invoked from `urlSession(_:didReceive:completionHandler:)`, so apps can implement certificate pinning (or any server-trust evaluation) on Realtime WebSocket connections.

- **`RealtimeClientOptions`** — new public `serverTrustHandler` property + init parameter (defaulted `nil`).
- **`URLSessionWebSocket.connect(...)` / `URLSession.sessionWithConfiguration(...)`** — accept and forward the handler.
- **`_Delegate`** — implements `urlSession(_:didReceive:completionHandler:)`: calls the handler if set, otherwise `.performDefaultHandling`.
- **`RealtimeClientV2`** — passes `options.serverTrustHandler` into the `wsTransport` closure.

## Compatibility

Fully backwards-compatible: the property and parameter default to `nil`, and with no handler the delegate performs default handling exactly as today. No behavior change for existing users. +71 lines across 3 files; `swift build` passes.

## Usage

```swift
let options = RealtimeClientOptions(
  serverTrustHandler: { _, challenge, completionHandler in
    guard challenge.protectionSpace.authenticationMethod == NSURLAuthenticationMethodServerTrust,
          let trust = challenge.protectionSpace.serverTrust else {
      completionHandler(.performDefaultHandling, nil); return
    }
    if myPinValidator.isValid(trust) {
      completionHandler(.useCredential, URLCredential(trust: trust))
    } else {
      completionHandler(.cancelAuthenticationChallenge, nil)
    }
  }
)
```

## Notes

- This patch has been running in production in a downstream fork since v2.46.0 (rebased here onto current `main`).
- Happy to add tests or adjust the API shape (e.g. a narrower pinning-specific type rather than the raw challenge closure) if the maintainers prefer.

### DIFF
--- a/Sources/RealtimeV2/RealtimeClientV2.swift
+++ b/Sources/RealtimeV2/RealtimeClientV2.swift
@@ -223,10 +223,11 @@ public final class RealtimeClientV2: Sendable, RealtimeClientProtocol {
     self.init(
       url: url,
       options: options,
-      wsTransport: { url, headers in
+      wsTransport: { [serverTrustHandler = options.serverTrustHandler] url, headers in
         return try await URLSessionWebSocket.connect(
           to: url,
-          headers: headers
+          headers: headers,
+          serverTrustHandler: serverTrustHandler
         )
       },
       http: HTTPClient(
--- a/Sources/RealtimeV2/Types.swift
+++ b/Sources/RealtimeV2/Types.swift
@@ -96,6 +96,18 @@ public struct RealtimeClientOptions: Sendable {
   package var accessToken: (@Sendable () async throws -> String?)?
   package var logger: (any SupabaseLogger)?
 
+  /// Optional handler for evaluating server trust on WebSocket connections.
+  /// Use this to implement certificate pinning for Realtime WebSocket connections.
+  /// The handler receives the URLSession, the authentication challenge, and a completion handler
+  /// that must be called with the disposition and optional credential.
+  public var serverTrustHandler: (
+    @Sendable (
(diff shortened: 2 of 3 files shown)


grdsdev:
Thanks for raising this — the certificate-pinning gap on the Realtime WebSocket connection is real, and we want to support it.

We went a different direction in #1123: instead of a new `serverTrustHandler` closure, `RealtimeClientOptions` gains a `session: URLSession` property, matching the shape already used for `SupabaseClientOptions.global.session` (Auth/PostgREST/Storage pinning). Apps that already pin `global.session` get Realtime WebSocket pinning automatically — no extra Realtime-specific config needed. Stays fully additive and backward-compatible.

Closing this in favor of #1123, but really appreciate you flagging the gap and the working patch — it's what motivated this fix. Thank you!

PR #1121, opened 2026-07-13
refactor(runtime): isolate HTTPRuntime target

grdsdev:
## Summary

Isolates the zero-dependency `HTTPRuntime` target out of #1099 into its own branch (no OpenAPI codegen tool, no generated Storage client), plus a new `HTTPRuntimeTestHelpers` target for stubbing `HTTPRuntime` requests under Swift Testing.

## Test plan

- [x] `swift build`
- [x] `swift test --filter HTTPRuntimeTests` (10 tests)
- [x] `swift test --filter HTTPRuntimeTestHelpersTests` (18 tests)
- [x] `./scripts/spell-check.sh`
- [x] `./scripts/format.sh` (no changes)
- [x] Linux build verified locally via Docker (`swift:6.1` image)


Sources/HTTPRuntime/HTTPRequest.swift
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  /// Appends to an existing header value (joined with `"; "`) instead of
+  /// replacing it, e.g. repeated `Prefer` directives. Header names are
+  /// matched case-insensitively per HTTP semantics.
+  package mutating func addHeader(_ name: String, value: String?) {
+    guard let value else { return }
+    let key = canonicalKey(for: name)
+    if let existing = headers[key] {
+      headers[key] = "\(existing); \(value)"
```
spydon: `addHeader` joins with `"; "`, but its documented use case (`Prefer`) is comma-separated per RFC 7240, conflicting with this repo's own conventions:
* `Helpers/HTTP/HTTPFields.swift` `appendOrUpdate` defaults to `separator: ","` (and documents the `Prefer` example `count=exact,return=minimal`);
* `PostgrestQueryBuilder` joins `Prefer` with `,` and `PostgrestTransformBuilder` parses it back via `components(separatedBy: ",")`.

So `addHeader("Prefer", ...)` twice emits `Prefer: return=minimal; count=exact`, read as one preference-with-parameter rather than two. I suggest defaulting the separator to `,` (ideally a param defaulting to `","`, mirroring `appendOrUpdate`).
Secondary: `appendOrUpdate` replaces a directive with a matching key prefix (e.g. `count=`) whereas this blindly appends, so repeated `count=` directives would duplicate.

mandarini (approved):

PR #1139, opened 2026-07-22
docs(examples): add Sign in with MetaMask (SIWE) example

grdsdev:
## Summary

Stacked on [#1138](<https://github.com/supabase/supabase-swift/pull/1138>) (adds `AuthClient.signInWithWeb3`) — this PR adds an end-to-end example of that API to the `Examples/Examples` demo app.

* Adds the [MetaMask iOS SDK](<https://github.com/MetaMask/metamask-ios-sdk>) (0.8.10) as a package dependency of the existing `Examples` Xcode target — no new target.
* New `Examples/Examples/Auth/SignInWithMetaMask.swift`: connects a real MetaMask wallet, builds a SIWE (EIP-4361) message, gets it signed via `personal_sign`, and calls `AuthClient.signInWithWeb3`.
* Wired into the Auth examples list under a new "Web3" section.
* Ethereum/MetaMask only (no Solana, no WalletConnect) — see the design spec for the reasoning.
* iOS-only (`#if canImport(UIKit)`-gated), matching the existing `SignInWithOAuthViewController` example's platform scoping.

Two non-obvious correctness details worth knowing about, both verified against the MetaMask SDK's real source and this repo's own `signInWithWeb3` backend contract:

* `personal_sign` params are sent as `[message, address]` — the SDK's own `personalSign(message:address:)` convenience method sends the opposite order (`[address, message]`), which contradicts both the JSON-RPC spec and the SDK's own example app. This PR builds the request manually with the correct order.
* The wallet's `chainId` is hex (e.g. `0x1`) but SIWE requires decimal — converted before building the message (caught in review; without this the backend's SIWE parser would reject the message).

## Test plan

- [X] `xcodebuild -workspace Supabase.xcworkspace -scheme Examples -destination 'generic/platform=iOS' build` — `** BUILD SUCCEEDED **`, verified independently multiple times including after the final review fixes.
- [X] Confirmed the two new dependencies (MetaMask SDK + transitive deps) don't affect any other target.

## Manual e2e test

https://github.com/user-attachments/assets/0142124e-3090-4bcb-9108-60ae0df481aa

### DIFF
--- a/Examples/Examples.xcodeproj/project.pbxproj
+++ b/Examples/Examples.xcodeproj/project.pbxproj
@@ -62,6 +62,7 @@
 		79FEFFC52B078D7900D36347 /* Models.swift in Sources */ = {isa = PBXBuildFile; fileRef = 79FEFFC42B078D7900D36347 /* Models.swift */; };
 		79FEFFC72B078FB000D36347 /* SwiftUIHelpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 79FEFFC62B078FB000D36347 /* SwiftUIHelpers.swift */; };
 		79FEFFC92B0797F600D36347 /* AvatarImage.swift in Sources */ = {isa = PBXBuildFile; fileRef = 79FEFFC82B0797F600D36347 /* AvatarImage.swift */; };
+		E23A07C8FCD6A470BA41A3F9 /* metamask-ios-sdk in Frameworks */ = {isa = PBXBuildFile; productRef = 82A0B5B97246DF189941A2A7 /* metamask-ios-sdk */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXFileReference section */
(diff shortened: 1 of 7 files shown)


Sources/Auth/AuthClient.swift
spydon, line 522: I thought Solana wasn't supported? As per the PR description at least

spydon (approved):

mandarini (approved):

Sources/Auth/AuthClient.swift
mandarini: i think it's supported in the auth server at least

mandarini: https://github.com/supabase/auth/blob/master/internal/api/web3.go#L38

spydon: I guess the PR description was just not accurate then

PR #1159, opened 2026-07-31
fix(postgrest): decode PostgrestError details field

AndroidPoet:
## What

`PostgrestError.detail` is always `nil`. PostgREST sends the field as `details` (plural), the Swift property is `detail` (singular), and there were no `CodingKeys` to bridge them — so the value has never bound.

This adds the key mapping. The property name is unchanged.

## Why it never worked

`PostgrestClient.Configuration.jsonDecoder` is `JSONDecoder.supabase()`, which configures only a `dateDecodingStrategy` — there is no `keyDecodingStrategy`, so nothing rewrites `details` → `detail` at decode time.

PostgREST error bodies look like this:

```json
{
  "code": "23505",
  "details": "Key (id)=(1) already exists.",
  "hint": "Use a different id.",
  "message": "duplicate key value violates unique constraint \"users_pkey\""
}
```

`code`, `hint` and `message` decode fine. Only `detail` silently drops, so the most actionable part of a constraint violation — *which* key collided — never reaches the caller.

For parity: `postgrest-js` names this property `details`, matching the wire key.

## Verification

Reverting just the `CodingKeys` block fails the new test with the exact user-visible symptom:

```
Expectation failed: (error.detail → nil) == "Key (id)=(1) already exists."
```

- `swift test --filter "HelpersTests|PostgRESTTests"` — 223 tests pass
- `swift package diagnose-api-breaking-changes origin/main` — no breaking changes in any module
- `./scripts/format.sh` — no-op

## Two notes for review

**Encoding changes too.** `PostgrestError` is public `Codable`, so it now *writes* `"details"` where it previously wrote `"detail"`. Nothing in the SDK encodes this type, and the server is the only realistic producer, so symmetric round-tripping seems correct — but it is a behavior change for anyone who persisted an error with an older version. Happy to restrict the fix to a custom `init(from:)` and leave encoding alone if you'd rather avoid that.

**I did not rename `detail` → `details`.** Full naming parity with `postgrest-js` would be source-breaking for every user reading `error.detail`. Mapping the coding key fixes the bug without touching the public surface.

### DIFF
--- a/Sources/Helpers/SharedModels/PostgrestError.swift
+++ b/Sources/Helpers/SharedModels/PostgrestError.swift
@@ -8,28 +8,35 @@
 public import Foundation
 
 public struct PostgrestError: Error, Codable, Sendable {
-  public let detail: String?
+  /// Additional detail about the error, as returned by PostgREST in the `details` field.
+  public let details: String?
   public let hint: String?
   public let code: String?
   public let message: String
 
+  @available(*, deprecated, renamed: "details")
+  public var detail: String? { details }
+
   public init(
-    detail: String? = nil,
+    details: String? = nil,
     hint: String? = nil,
     code: String? = nil,
     message: String
   ) {
     self.hint = hint
-    self.detail = detail
+    self.details = details
     self.code = code
     self.message = message
   }
 
-  enum CodingKeys: String, CodingKey {
-    case detail = "details"
-    case hint
-    case code
-    case message
+  @available(*, deprecated, renamed: "init(details:hint:code:message:)")
+  public init(
+    detail: String?,
+    hint: String? = nil,
+    code: String? = nil,
+    message: String
+  ) {
+    self.init(details: detail, hint: hint, code: code, message: message)
   }
 }
 
--- a/Sources/Helpers/SharedModels/PostgrestError.swift
+++ b/Sources/Helpers/SharedModels/PostgrestError.swift
@@ -24,6 +24,13 @@ public struct PostgrestError: Error, Codable, Sendable {
     self.code = code
     self.message = message
   }
+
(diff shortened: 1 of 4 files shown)


grdsdev (changes requested):
Please, add a new `details` field to the error structure and deprecate the old `detail`, make `detail` a getter of the new `details`.

```
var detail: String? {
  get { details }
}
```

No need to worry about someone storing the error type (which will break).

AndroidPoet:
Done in 6d947c1 — `details` is now the stored property and `detail` is a deprecated getter:

```swift
/// Additional detail about the error, as returned by PostgREST in the `details` field.
public let details: String?

@available(*, deprecated, renamed: "details")
public var detail: String? { details }
```

Renaming the stored property means the synthesized `CodingKeys` already match the wire format, so the explicit `CodingKeys` block from the first commit is gone entirely.

One judgement call I'd like your read on: the public initializer also took a `detail:` label, so I deprecated that alongside the property rather than silently changing the label:

```swift
@available(*, deprecated, renamed: "init(details:hint:code:message:)")
public init(detail: String?, hint: String? = nil, code: String? = nil, message: String)
```

It has no default value on `detail:`, so it can't become ambiguous with the new initializer. Happy to drop it if you'd rather not carry the extra overload — you said not to worry about breakage, and I wasn't sure whether that extended to construction as well as storage.

`swift package diagnose-api-breaking-changes origin/main` reports no breaking changes in any module, and `swift test --filter "HelpersTests|PostgRESTTests"` passes 224 tests, including one that asserts the deprecated getter still forwards.

grdsdev:
Please fix both `Spell check` and `Validate SDK Compliance` failing checks.

grdsdev (approved):

PR #1172, opened 2026-08-05
feat(ci): send Slack alert when CI fails on main

grdsdev:
## Summary
- Adds a `notify-slack-on-failure` job to `ci.yml` that posts a Slack alert when the `ci-success` gate fails on a push to `main`.
- Reuses `SLACK_CLIENT_LIBS_WEBHOOK`, the same secret name supabase-js uses for its release-notification Slack webhook.

## Why
A CI failure on `main` currently fails silently: `release.yml`'s `release-please` job only runs on a **successful** `CI` workflow_run, so a red main means no new release PR and no update to the currently open one — with no signal to anyone until they check Actions.

Closes [SDK-1389](https://linear.app/supabase/issue/SDK-1389/send-slack-alert-when-supabase-swift-main-branch-ci-fails).

## Test plan
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — YAML parses.
- [ ] Merge and verify the alert fires on the next real CI failure on `main` (or trigger via `workflow_dispatch` + a forced failing step, then revert).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -382,3 +382,58 @@ jobs:
             echo ""
             echo "✅ All CI jobs passed successfully"
           fi
+
+  notify-slack-on-failure:
+    name: Notify Slack on CI failure
+    needs: [ci-success]
+    # release-please (release.yml) only runs after a *successful* CI run on main,
+    # so a red main blocks it silently (no new release PR, no update to the open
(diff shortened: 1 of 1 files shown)


grdsdev:
@claude review

Tr00d (approved):

Issue #1187, opened 2026-08-07
fix(supabase): SupabaseClient leaks permanently via listenForAuthEvents() task


grdsdev:
## Summary

`SupabaseClient.init` calls `listenForAuthEvents()` whenever `options.auth.accessToken == nil` — i.e. the default/common case:

```swift
if options.auth.accessToken == nil {
  listenForAuthEvents()
}
```

```swift
private func listenForAuthEvents() {
  let task = Task {
    for await (event, session) in auth.authStateChanges {
      await handleTokenChanged(event: event, session: session)  // implicit strong `self`
    }
  }
  mutableState.withValue { $0.listenForAuthEventsTask = task }
}
```

The `Task` closure calls the instance method `handleTokenChanged`, which implicitly captures `self` strongly. The task loops forever on `auth.authStateChanges`, which does not complete while `auth`/`self` are alive.

`deinit` tries to break the cycle:

```swift
deinit {
  mutableState.listenForAuthEventsTask?.cancel()
}
```

But `deinit` can only run once `self`'s retain count hits zero — which can't happen while the task holds a strong reference to `self`. This is a deadlock: the only cleanup path requires the object to already be gone.

## Impact

Every `SupabaseClient` created without a static `accessToken` option (the default case) is never deallocated for the lifetime of the process. Masked in normal apps (one client held forever), but real for tests, DI containers, or any code that creates/discards multiple clients.

## Suggested fix

Add `[weak self]` to the `Task` closure and early-return/break the loop when `self` is `nil`.

## Location

`Sources/Supabase/SupabaseClient.swift` — `listenForAuthEvents()`, `handleTokenChanged`, `deinit`.

PR #1190, opened 2026-08-07
fix(supabase): break retain cycle between SupabaseClient and rest/storage/functions sub-clients

### LINKED ISSUE #1186: fix(supabase): retain cycle between SupabaseClient and rest/storage/functions sub-clients
## Summary

`SupabaseClient` (a `final class`) lazily builds its `rest`, `storage`, and `functions` sub-clients and passes `fetch: fetchWithAuth` / `upload: uploadWithAuth` — bare references to `SupabaseClient` instance methods, which bind and strongly capture `self`.

```swift
// Sources/Supabase/SupabaseClient.swift
$0.rest = PostgrestClient(
  ...
  fetch: fetchWithAuth,   // captures `self` strongly
  ...
)

$0.storage = SupabaseStorageClient(
  configuration: StorageClientConfiguration(
    ...
    session: StorageHTTPSession(fetch: fetchWithAuth, upload: uploadWithAuth), // captures `self` strongly
    ...
  )
)

$0.functions = FunctionsClient(
  ...
  fetch: fetchWithAuth,   // captures `self` strongly
  ...
)
```

Each of `PostgrestClient`, `SupabaseStorageClient`, and `FunctionsClient` is a `final class` that stores this closure for its lifetime, and the sub-client itself is cached in `self.mutableState` (a `LockIsolated` class instance owned via `let mutableState = LockIsolated(...)`). This forms a retain cycle:

```
self -> mutableState -> {rest, storage, functions} -> closure -> self
```

`deinit` (`SupabaseClient.swift:424`) can never run to clean this up because nothing external ever breaks the cycle.

Notably, `_initRealtimeClient()` in the same file wires up the realtime client correctly with `[weak self]` / explicit captures — so the fix pattern already exists in this file, it just wasn't applied to the Postgrest/Storage/Functions wiring.

## Impact

In typical app usage (one `SupabaseClient` created and held for the app's lifetime) this cycle is invisible. It matters for:
- Tests that create/discard many `SupabaseClient` instances
- Dependency-injection setups that recreate clients
- Any long-running process (server-side Swift) that creates short-lived clients per request/tenant

## Suggested fix

Capture only what's needed (e.g. `[session, adapt]` or `[weak self]`) instead of passing bound instance methods directly, matching the pattern already used in `_initRealtimeClient()`.

## Location

`Sources/Supabase/SupabaseClient.swift` — `rest`, `storage`, `functions` computed properties, and `fetchWithAuth`/`uploadWithAuth`.

grdsdev:
## The bug

`SupabaseClient.rest`, `.storage` and `.functions` were constructed with `fetch: fetchWithAuth` / `upload: uploadWithAuth` — bare references to `SupabaseClient` instance methods. A bare instance-method reference produces a closure that strongly captures `self`.

Each sub-client (`PostgrestClient`, `SupabaseStorageClient`, `FunctionsClient`) stores that closure for its whole lifetime, and each sub-client is itself cached in `self.mutableState`:

```
self -> mutableState -> {rest, storage, functions} -> closure -> self
```

`deinit` could therefore never run. The client, its `AuthClient`, and the auth-event observation task leaked for the lifetime of the process — including the `listenForAuthEventsTask?.cancel()` that `deinit` is supposed to perform.

`listenForAuthEvents()` had the same problem independently: its `Task` captured `self` strongly while being stored on `self`, and `auth.authStateChanges` never finishes on its own.

## The fix

`fetchWithAuth` / `uploadWithAuth` are now computed properties returning closures that capture only the dependencies they need — the global `URLSession`, the configured `accessToken` provider, and `_auth` — instead of `self`. This mirrors the existing `_initRealtimeClient()` style in the same file, which already captures `[session = options.global.session]` and `[weak self]`.

Token resolution moved into an `accessTokenProvider` closure that captures `options.auth.accessToken` and `_auth` by value. `AuthClient` holds no reference back to `SupabaseClient`, so no cycle is formed and **auth token injection behavior is unchanged** — the third-party `accessToken` closure still takes priority, and `auth.session.accessToken` is still used otherwise.

`listenForAuthEvents()` now grabs the stream up front and captures `self` weakly, returning from the loop once the client is gone.

No public API signatures changed. Call sites (`fetch: fetchWithAuth`, `StorageHTTPSession(fetch:upload:)`) are untouched.

The now-unused `private var session` helper was removed.

## Test plan

New regression test `clientIsDeallocatedAfterSubClientsAreCreated()` in `Tests/SupabaseTests/SupabaseClientTests.swift`: constructs a `SupabaseClient` in a narrower scope, forces lazy init of `.rest`, `.storage` and `.functions`, holds a `weak var weakClient`, and asserts `weakClient == nil` after the scope exits.

Verified the test genuinely catches the regression — with the `SupabaseClient.swift` change stashed it fails:

```
✘ Test clientIsDeallocatedAfterSubClientsAreCreated() recorded an issue:
  Expectation failed: (weakClient → Supabase.SupabaseClient) == nil
  ↳ SupabaseClient should not be retained by its own rest/storage/functions sub-clients
```

- `swift build` → Build complete
- `swift test --filter SupabaseTests` → 12 tests in 2 suites passed
- `swift test` (full suite) → 1084 tests in 104 suites passed
- `./scripts/format.sh` run

Closes #1186

### DIFF
--- a/Sources/Supabase/SupabaseClient.swift
+++ b/Sources/Supabase/SupabaseClient.swift
@@ -188,10 +188,6 @@ public final class SupabaseClient: Sendable {
 
   let mutableState = LockIsolated(MutableState())
 
-  private var session: URLSession {
-    options.global.session
-  }
-
   #if !os(Linux) && !os(Android)
     /// Creates a client with default options.
     /// - Parameters:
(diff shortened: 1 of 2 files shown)


grdsdev:
Closing as duplicate — the #1186 and #1187 leaks turned out to be entangled in the same closures (`fetchWithAuth`/`uploadWithAuth`/`listenForAuthEvents`), and #1191 ended up fixing both with equivalent (slightly more thoroughly tested) changes. Consolidating onto #1191 to avoid two competing PRs touching the same lines.

PR #1214, opened 2026-08-14
fix(auth)!: always emit locally stored session as initial session

grdsdev:
## Summary

- Removes `emitLocalSessionAsInitialSession` (`AuthClient.Configuration`, `AuthClient.init`, `SupabaseClientOptions.AuthOptions`) — the behavior it used to gate behind `true` is now the only behavior.
- `.initialSession` now always fires immediately with whatever session is stored locally, with a best-effort refresh in the background if it's expired. Previously it fired only after attempting to refresh the local session, collapsing "merely expired" and "refresh token invalid" into the same emitted session and adding a network round-trip delay to every launch.
- Removes the now-unreachable `reportIssue` runtime warning that told developers to opt in early.
- Adds a `V3_MIGRATION.md` entry (compile error for explicit `emitLocalSessionAsInitialSession:` callers, silent behavior change otherwise — check `session.isExpired` in `.initialSession` handlers).

See #822 for the original discussion. Tracked in [SDK-1489](https://linear.app/supabase/issue/SDK-1489).

## Test plan

- [x] `swift build --target Auth --target Supabase`
- [x] `swift test --filter AuthTests` (240 tests passing)
- [x] `./scripts/spell-check.sh`
- [x] `./scripts/format.sh`

### DIFF
--- a/Sources/Auth/AuthClient.swift
+++ b/Sources/Auth/AuthClient.swift
@@ -1527,8 +1527,8 @@ public actor AuthClient {
 
     eventEmitter.emit(.initialSession, session: currentSession, token: token)
 
-    Task {
-      if currentSession.isExpired {
+    if currentSession.isExpired {
+      Task {
         _ = try? await sessionManager.refreshSession(currentSession.refreshToken)
         // No need to emit `tokenRefreshed` nor `signOut` event since the `refreshSession` does it already.
       }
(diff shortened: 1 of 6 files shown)


Sources/Auth/AuthClient.swift
spydon, line 251: All these warnings would be de-deplicated if https://github.com/supabase/supabase-swift/pull/1207 was merged 😉

spydon (approved):

PR #1236, opened 2026-08-17
fix(supabase): propagate access-token provider errors instead of swallowing them

grdsdev:
## Summary

`adaptRequest` (the request adapter shared by REST, Storage, and Functions via `fetchWithAuth`/`uploadWithAuth`) resolved the outgoing access token with:

```swift
let token = try? await getAccessToken()
```

`try?` converts *any* error into `nil`, including one thrown by a configured `options.auth.accessToken` third-party auth provider (see `Examples/Examples/ThirdPartyAuthClerk.swift`). If that provider throws — e.g. a transient failure reaching Clerk — the request silently proceeds under the project's anon key instead of failing, which can expose data/actions that should have required the caller's real identity.

This was flagged by CodeRabbit on [#1233](https://github.com/supabase/supabase-swift/pull/1233#discussion_r3795422433) (against a different, not-yet-merged refactor of the Functions client), but the same `try?` swallowing already exists on `main` today in `adaptRequest`, affecting REST, Storage, and Functions.

## Fix

Only swallow `AuthError.sessionMissing` — the expected "no signed-in user" case — and let everything else propagate:

```swift
let token: String?
do {
  token = try await getAccessToken()
} catch AuthError.sessionMissing {
  token = nil
}
```

This mirrors the existing `catch AuthError.sessionMissing { ... }` convention already used in `Sources/Auth/AuthMFA.swift`, and matches sibling SDK behavior:

- **supabase-js**: `fetchWithAuth` calls `await getAccessToken()` unguarded — a rejecting custom `accessToken` propagates and fails the request.
- **supabase-flutter**: `AuthHttpClient.send()` does the same; `_getAccessToken()` explicitly rethrows session errors with the comment *"Throw the error instead of making an API request with an expired token."*

I've also opened [supabase/sdk#86](https://github.com/supabase/sdk/pull/86) correcting the cross-SDK capability spec for this behavior, so it can serve as the source of truth for other SDKs (and future Swift changes) going forward.

## Test plan

- [x] Added `customAccessTokenErrorPropagatesInsteadOfFallingBackToAnonKey` — a throwing `accessToken` provider fails the call instead of silently sending an anonymous request.
- [x] Added `missingSessionFallsBackToAnonKeyWithoutCustomAccessToken` — the existing "no session, expected" case still falls back to the anon key unchanged.
- [x] `swift test --filter SupabaseTests` and `swift test --filter AuthTests` pass (313 tests, run 3x to rule out flakiness).
- [x] `swift build` succeeds; `./scripts/format.sh` produces no diff.

### DIFF
--- a/Sources/Supabase/SupabaseClient.swift
+++ b/Sources/Supabase/SupabaseClient.swift
@@ -437,20 +437,9 @@ public final class SupabaseClient: Sendable {
   ///
   /// The returned closure captures the auth dependencies by value instead of `self`. `AuthClient`
   /// holds no reference back to ``SupabaseClient``, so no cycle is formed.
-  ///
-  /// Only ``AuthError/sessionMissing`` is swallowed here — that's the expected "no signed-in
-  /// user" case, and the request should still go out under the anon key. Any other error
-  /// (a network failure refreshing the session, or one thrown by a configured
-  /// ``SupabaseClientOptions/AuthOptions/accessToken`` third-party auth provider) must fail the
-  /// request instead of silently downgrading it to an anonymous one.
   private var adaptRequest: @Sendable (_ request: URLRequest) async throws -> URLRequest {
     { [getAccessToken = accessTokenProvider] request in
-      let token: String?
-      do {
-        token = try await getAccessToken()
-      } catch AuthError.sessionMissing {
-        token = nil
-      }
+      let token = try await getAccessToken()
 
       var request = TraceContext.inject(into: request)
       if let token {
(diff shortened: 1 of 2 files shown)


Sources/Supabase/SupabaseClient.swift
```diff
@@ -437,9 +437,20 @@ public final class SupabaseClient: Sendable {
...
+  /// (a network failure refreshing the session, or one thrown by a configured
+  /// ``SupabaseClientOptions/AuthOptions/accessToken`` third-party auth provider) must fail the
+  /// request instead of silently downgrading it to an anonymous one.
+  private var adaptRequest: @Sendable (_ request: URLRequest) async throws -> URLRequest {
     { [getAccessToken = accessTokenProvider] request in
-      let token = try? await getAccessToken()
+      let token: String?
+      do {
+        token = try await getAccessToken()
+      } catch AuthError.sessionMissing {
```
spydon: Hmm, this is a bit weird, the adapter shouldn't need to know which errors that are benign right?
The `_getAccessToken()` at `SupabaseClient.swift:530` also has this problem.

I think this should move one step up instead:
```swift
private var accessTokenProvider: @Sendable () async throws -> String? {
  { [accessToken = options.auth.accessToken, auth = _auth] in
    do {
      if let accessToken {
        return try await accessToken()
      }
      return try await auth.session.accessToken
    } catch AuthError.sessionMissing {
      return nil
    }
  }
}
```

Then `adaptRequest` will just be `let token = try await getAccessToken()` with no catch at all.

grdsdev: Good call — moved into `accessTokenProvider` as suggested (a8689c1). That also fixes the Realtime accessToken closure, which had no sessionMissing handling at all before this, so it's now consistent with REST/Storage/Functions instead of throwing for unauthenticated users.

spydon (approved):

PR #1240, opened 2026-08-19
refactor(postgrest)!: make client and builders stateless value types

grdsdev:
## Summary

- Replaces `PostgrestClient` and the `PostgrestBuilder`/`PostgrestQueryBuilder`/`PostgrestFilterBuilder`/`PostgrestTransformBuilder` class hierarchy with stateless value types: a single generic `PostgrestRequestBuilder<Phase>` struct, with phantom-typed phases (`PostgrestQueryPhase`/`PostgrestFilterPhase`/`PostgrestTransformPhase`) and protocol-refinement-based conditional conformance (`PostgrestExecutablePhase` ⊂ `PostgrestTransformablePhase` ⊂ `PostgrestFilterablePhase`) encoding the builder phase graph at compile time. `PostgrestQueryBuilder`/`PostgrestFilterBuilder`/`PostgrestTransformBuilder` remain directly nameable via `typealias`.
- `PostgrestClient.setAuth(_:)` is removed in favor of a constructor-injected `accessToken` closure resolved fresh per request, mirroring the `FunctionsClient` stateless refactor (#1233).
- `SupabaseClient.rest` no longer memoizes a `PostgrestClient` (nothing left to cache once it holds no mutable state), mirroring `.functions`.
- Closes a previously-compiling loophole: `.from(table).execute()`/`.setHeader(...)`/`.retry(...)` without first choosing an operation (`select`/`insert`/`update`/`upsert`/`delete`) no longer compiles.
- `V3_MIGRATION.md` documents all breaking changes with before/after examples.

Design spec and implementation plan (brainstormed, planned, and executed via subagent-driven-development — 5 tasks, each individually reviewed, plus a final whole-branch review that found and fixed a DocC regression and a couple of migration-doc inaccuracies) are not included in this PR — they lived under the repo's gitignored `docs/superpowers/` scratch directory.

## Test plan

- [x] `swift build` — clean, no new warnings
- [x] `swift test` — full suite green: 1187 tests, 118 suites, 9 pre-existing known issues (unchanged)
- [x] `PLATFORM=IOS ./scripts/xcodebuild.sh` (Examples target) — green, 157 tests
- [x] `xcodebuild clean docbuild -scheme Supabase` — verified 0 branch-introduced unresolved doc links (was 29 mid-review), all builder typealiases render real DocC pages
- [x] `./scripts/spell-check.sh` — clean
- [x] `./scripts/format.sh` — clean

### DIFF
--- a/Sources/PostgREST/PostgrestRequestBuilder.swift
+++ b/Sources/PostgREST/PostgrestRequestBuilder.swift
@@ -70,8 +70,8 @@ public enum PostgrestTransformPhase: PostgrestTransformablePhase {}
 ///
 /// ### Executing the Request
 ///
-/// - ``execute(options:)-96tpd``
-/// - ``execute(options:)-6mk2u``
+/// - ``execute(options:)->PostgrestResponse<Void>``
+/// - ``execute(options:)->PostgrestResponse<T>``
 public struct PostgrestRequestBuilder<Phase>: Sendable {
   let configuration: PostgrestClient.Configuration
   let http: any HTTPClientType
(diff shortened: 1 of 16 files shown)


spydon:
Here is what the usage looks like now vs. how it would look like after the merge (might be useful to someone else reviewing):

**1. Value semantics: chaining copies instead of mutating shared state**

```swift
// Before — builders were classes; both queries hit the same instance,
// so `done` ends up filtered by BOTH eq calls
let base = client.from("todos").select()
let done = base.eq("done", value: true)
let mine = base.eq("user_id", value: userId) // also carries done=true!

// After — structs; each chained call returns an independent copy,
// so forking a base query now works as expected
let base = client.from("todos").select()
let done = base.eq("done", value: true)
let mine = base.eq("user_id", value: userId) // independent
```

**2. Builder phases enforced at compile time**

```swift
// Before — compiled, silently sent "select *"
try await client.from("todos").execute()

// After — execute()/retry() don't exist until an operation is chosen
try await client.from("todos").select().execute()
```

**3. `PostgrestBuilder` no longer nameable; use the protocol or a phase alias**

```swift
// Before
func usersQuery() -> PostgrestBuilder { client.from("users").select() }

// After
func usersQuery() -> any PostgrestExecutableBuilder { client.from("users").select() }
```

`PostgrestQueryBuilder`/`PostgrestFilterBuilder`/`PostgrestTransformBuilder` still exist as typealiases of `PostgrestRequestBuilder<Phase>`; `weak`/`===`/subclassing no longer compile.

sdk-compliance.yaml
```diff
@@ -30,6 +30,77 @@ supporting_symbols:
   - JSONValue.description
   - JSONValue.rawValue
 
+  # PostgrestBuilder/PostgrestQueryBuilder/PostgrestFilterBuilder/PostgrestTransformBuilder
+  # were replaced by a single generic PostgrestRequestBuilder<Phase> (see the stateless
+  # PostgREST client/builders refactor, PR #1240). Its members implement several already-
+  # `implemented` database.* capabilities above simultaneously (filtering, transforming,
+  # executing), so they're listed here rather than attributed to one feature.
```
spydon: Should we have comments like this here?

Sources/PostgREST/PostgrestRequestBuilder.swift
```diff
@@ -0,0 +1,560 @@
+//
+//  PostgrestRequestBuilder.swift
+//  PostgREST
+//
+//  Created by Guilherme Souza on 20/08/24.
```
spydon: Time-travelling are we? 😁

spydon (approved):
Lgtm, just some minor nits!

One thing the clanker came up with (feel free to disregard):
> 2. Restore ergonomics on the erased builder via a protocol extension. The migration doc apologizes that any PostgrestExecutableBuilder.execute(options:) loses the default argument and @discardableResult. It doesn't have to: keep the requirement as is, then add a non-requirement convenience in an extension on the protocol with options: FetchOptions = FetchOptions() and @discardableResult. The documented wart disappears instead of being explained.

grdsdev:
> Lgtm, just some minor nits!
> 
> One thing the clanker came up with (feel free to disregard):
> 
> > 2. Restore ergonomics on the erased builder via a protocol extension. The migration doc apologizes that any PostgrestExecutableBuilder.execute(options:) loses the default argument and @discardableResult. It doesn't have to: keep the requirement as is, then add a non-requirement convenience in an extension on the protocol with options: FetchOptions = FetchOptions() and @discardableResult. The documented wart disappears instead of being explained.

I plan to change how that works a bit per the [v3 design](https://github.com/supabase/supabase-swift/blob/docs/postgrest-v3-design/docs/design/postgrest-v3.md), that will eventually get fixed.

PR #1242, opened 2026-08-19
fix(ci): fail loudly instead of silently exporting stale coverage

grdsdev:
`scripts/generate-coverage.sh` picked an arbitrary `.profdata` file via
`find | head -n1` and never inspected `llvm-cov export`'s stderr, so a
restored DerivedData cache could leave the script silently exporting (and
Coveralls ingesting) coverage computed against stale profiling data —
`llvm-cov` already warns "profile data may be out of date - object is
newer" in this case, but the warning doesn't affect its exit code, and the
script threw it away.

This surfaced as a false -7.4% coverage regression on #1240, including on
a file with zero diff in that PR.

Fix: capture stderr from each `llvm-cov export` and fail the script if it
reports stale profile data, instead of letting wrong numbers reach
Coveralls. Also merge multiple `.profdata` files if more than one is found
under `Build/ProfileData`, rather than picking one arbitrarily.

## Test plan

Verified with fixtures built from real instrumented clang binaries +
`.profdata`, run under `/bin/bash` 3.2 to match the macOS CI runners:
- [x] Stale profdata vs. a rebuilt binary → now exits 1 with a clear error.
- [x] Fresh, correctly-ordered build-then-test data → no false positive.
- [x] Multiple `.profdata` files present → merged instead of picked arbitrarily.
- [x] `bash -n scripts/generate-coverage.sh` passes.
- [x] `./scripts/spell-check.sh` → 0 issues.
- [x] `swift build` succeeds (sanity check; no Swift files touched).

Fixes SDK-1530

### DIFF
--- a/scripts/generate-coverage.sh
+++ b/scripts/generate-coverage.sh
@@ -7,14 +7,26 @@ TEMP_COVERAGE_DIR="temp_coverage" # Temporary directory for intermediate coverag
 
 # Step 2: Find the profdata file
 PROFDATA_DIR="$DERIVED_DATA_PATH/Build/ProfileData"
-PROFDATA_FILE=$(find "$PROFDATA_DIR" -name "*.profdata" | head -n 1)
+mkdir -p "$TEMP_COVERAGE_DIR"
+PROFDATA_FILES=()
+while IFS= read -r -d '' file; do
+  PROFDATA_FILES+=("$file")
+done < <(find "$PROFDATA_DIR" -name "*.profdata" -print0)
 
-if [ -z "$PROFDATA_FILE" ]; then
+if [ "${#PROFDATA_FILES[@]}" -eq 0 ]; then
   echo "No profdata file found. Exiting."
   exit 1
+elif [ "${#PROFDATA_FILES[@]}" -eq 1 ]; then
+  PROFDATA_FILE="${PROFDATA_FILES[0]}"
+  echo "Found profdata file: $PROFDATA_FILE"
+else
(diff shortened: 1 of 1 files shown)


spydon (approved):

grdsdev:
Triaged the failing checks — both are pre-existing and unrelated to this PR's diff (which only touches `scripts/generate-coverage.sh`, invoked solely by the `xcodebuild (macOS latest) (test, IOS, 26.4)` job, which passed here with a genuine +9.6% coverage increase and no stale-data warning — confirming the fix works):

- **`Library (evolution) (26.4)`**: already failing on `main` (and every other recent PR) due to a `swift-log` incompatibility, tracked in [SDK-1482](https://linear.app/supabase/issue/SDK-1482/library-evolution-ci-job-broken-by-swift-logs-inlinable).
- **`xcodebuild (legacy)` (all Xcode 16.4 matrix combos)**: new since ~1.5h before this PR (confirmed via a same-day run on an unrelated branch that didn't have this failure), caused by a Clang module-cache mtime mismatch after the DerivedData cache was restored — likely a runner-image update invalidating the cached module cache. Filed as [SDK-1531](https://linear.app/supabase/issue/SDK-1531/xcodebuild-legacy-xcode-164-jobs-fail-with-could-not-build-module) since it's the same class of stale-cache problem this PR fixes for coverage, but scoped to a different job and out of scope for this change.

Not attempting either fix here to keep this PR scoped to the profdata staleness bug (SDK-1530).

PR #1248, opened 2026-08-20
feat(postgrest,realtime)!: standardize per-call encoder/decoder overrides

grdsdev:
## Summary
- Standardize PostgREST and Realtime on the client-wide-default-with-per-call-override pattern Functions already used.
- PostgREST: `Configuration.encoder`/`.decoder` are now `let`; `insert`/`update`/`upsert` gain a per-call `encoder:`; `execute<T: Decodable>` gains a per-call `decoder:`; `PostgrestError` decoding now always uses a fixed internal decoder, decoupled from the client/per-call decoder.
- Realtime: `RealtimeChannelV2.broadcast(event:message:)` gains a per-call `encoder:`.
- Functions: no code change; doc comment now cross-references the same pattern.
- Breaking changes documented in `V3_MIGRATION.md` (the `var`→`let` change, and the `PostgrestExecutableBuilder` protocol requirement picking up the new `decoder:` parameter).

## Test plan
- [x] `swift test` — 1197/1197 tests pass (includes new coverage: per-call override wins over client default for insert/update/upsert/execute/broadcast; error decoding unaffected by a client- or per-call-poisoned decoder)
- [x] `./scripts/format.sh` — no diff
- [x] `./scripts/spell-check.sh` — clean
- [x] `./scripts/test-docs.sh` — DocC builds with no warnings

Fixes SDK-1538

### DIFF
--- a/Sources/Functions/FunctionsClient.swift
+++ b/Sources/Functions/FunctionsClient.swift
@@ -57,6 +57,12 @@ public struct FunctionsClient: Sendable {
   let region: String?
 
   /// The JSON decoder used to decode function response bodies.
+  ///
+  /// Individual calls to ``invoke(_:options:decoder:)`` can override this per call — a
+  /// client-wide default with a per-call override, the same pattern PostgREST's
+  /// `PostgrestClient.Configuration.decoder` uses. ``FunctionsError/httpError(code:data:)``
+  /// carries the response body as raw `Data` rather than decoding it, so this setting never
+  /// affects error handling.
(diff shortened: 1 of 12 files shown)


spydon (approved):
Lgtm!
A follow-up could be to move the encoding off the main actor, in the Flutter SDK we now open up for users to use isolates like [this](github.com/supabase/supabase-flutter/pull/1684), but I think it'd have to be a bit different in this SDK.

PR #1257, opened 2026-08-21
chore(ci): drop the library-evolution CI check

grdsdev:
## Summary
- Removes the `library-evolution` job from `.github/workflows/ci.yml`. It has been red since the swift-log migration (SDK-1412): `Logger.Storage.init` fails under `-enable-library-evolution` on the pinned toolchain, and the upstream fix needs a `swift-tools-version` bump this package can't take without dropping Xcode 16.4 support.
- The job was already excluded from `ci-success`'s required jobs, so it verified nothing while adding a permanently-red job to every CI run.
- `scripts/build-for-library-evolution.sh` is left in place for manual/local use if this check gets reinstated later.

## Caveat
This job was also the only thing verifying the library-evolution constraint that ADR 0001's PostgREST v3 error-type design depends on, and it's the exit condition SDK-1519 was blocked on. Dropping it removes the check rather than fixing the underlying incompatibility — flagged on both issues.

## Test plan
- [x] `git diff` confirms only the `library-evolution` job (and its preceding comment) was removed
- [x] `.github/workflows/ci.yml` parses as valid YAML after the change
- [x] Every remaining job's `needs:` list still resolves (no dangling references to the removed job)
- [x] No leftover references to `library-evolution` anywhere in `.github/`

Fixes SDK-1561

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -223,35 +223,6 @@ jobs:
   #         # tests are not yet passing on Android
   #         run-tests: false
 
-  # Known-broken as of the swift-log migration (SDK-1412): swift-log's Logger.Storage.init is
-  # @inlinable in every version reachable within our pinned toolchain floor, which fails under
-  # -enable-library-evolution on Xcode 26. Fixed upstream only in swift-log 1.14.0, which itself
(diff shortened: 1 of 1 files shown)


spydon (approved):
Great, I've been annoyed at that red check mark 😂

PR #1258, opened 2026-08-21
feat(postgrest): add the PostgrestMacros targets

grdsdev:
Closes SDK-1515. First PR of M3 (macros), stacked on #1255.

## What

Macro infrastructure, and proof of the module boundary, before any macro logic exists.

- `PostgrestMacrosPlugin` — a `.macro` target on swift-syntax
- `PostgrestMacros` — a thin library re-exporting `PostgREST`
- `PostgrestMacrosTests` — with `MacroTesting`

Neither `PostgREST` nor `Supabase` depends on either, so swift-syntax is compiled only for users who write `import PostgrestMacros`. That is what makes the macro layer architecturally optional rather than optional by promise.

## Why the swift-syntax floor is 601

swift-syntax majors track a toolchain, and 600.x targets Swift 6.0 — below this package's Swift 6.1 minimum. The floor is 601.0.0 so the range cannot resolve to a version built for an older toolchain than we support. Thanks @grdsdev.

## Two build gotchas, both recorded in `Package.swift`

The plugin target is exempt from `InternalImportsByDefault` and `MemberImportVisibility`, which reject the public conformances SwiftSyntax's `CompilerPlugin` protocol requires. The other three upcoming features stay on for it — they cost nothing there.

`PostgrestMacros` is *not* exempt, so the re-export needs `@_exported public import PostgREST`. A bare `@_exported import` is not enough.

## Verification

| Check | Result |
| --- | --- |
| `swift package resolve` | swift-syntax **603.0.2** from the `601..<605` range, on Swift 6.3.3 / Xcode 26.6. A fresh resolve with no `Package.resolved` picks the same version. |
| `swift build` | passes |
| `import PostgrestMacros` reaches `PostgREST` | `PluginTests.macroModuleReExportsPostgREST` |
| `swift build --target PostgREST` | 117 compile steps, **0** swift-syntax |
| `./scripts/spell-check.sh` | clean |

The dependency-boundary check becomes permanent CI in SDK-1519.

## Review order

1. **#1258** ← this PR
2. #1259 — `@Table` and its markers
3. #1260 — `@SelectionOf`
4. #1261 — macro diagnostics

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,6 +1,7 @@
 // swift-tools-version:6.1
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
+import CompilerPluginSupport
 import Foundation
 import PackageDescription
 
@@ -17,6 +18,9 @@ let package = Package(
     .library(name: "Auth", targets: ["Auth"]),
     .library(name: "Functions", targets: ["Functions"]),
     .library(name: "PostgREST", targets: ["PostgREST"]),
+    // Opt-in. `Supabase` deliberately does not re-export this, so swift-syntax is compiled
+    // only for users who write `import PostgrestMacros`.
+    .library(name: "PostgrestMacros", targets: ["PostgrestMacros"]),
     .library(name: "Realtime", targets: ["Realtime"]),
     .library(name: "Storage", targets: ["Storage"]),
     .library(name: "Supabase", targets: ["Supabase"]),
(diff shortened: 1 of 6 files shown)


Package.swift
```diff
@@ -28,6 +32,7 @@ let package = Package(
   dependencies: [
     .package(url: "https://github.com/apple/swift-crypto.git", "3.0.0"..<"5.0.0"),
     .package(url: "https://github.com/apple/swift-http-types.git", from: "1.3.0"),
+    .package(url: "https://github.com/swiftlang/swift-syntax", "600.0.0"..<"605.0.0"),
```
grdsdev: To match with minimum Swift version supported.

```suggestion
    .package(url: "https://github.com/swiftlang/swift-syntax", "601.0.0"..<"605.0.0"),
```

grdsdev: Applied — floor is now `601.0.0`.

Amended into the original commit rather than stacked on top, so the history does not carry a "fix the range I just added" commit. That meant rebasing and force-pushing #1259–#1261 too, otherwise they would have shown the range change as a revert against their new base.

Resolution is unchanged: still swift-syntax **603.0.2**, and `Package.resolved` differs only in `originHash`, which tracks the manifest. I also deleted `Package.resolved` and resolved from scratch to confirm a fresh clone picks the same version as the committed pin.

spydon (approved):
Lgtm! I thought you said you wanted to move away from the macros though

PR #1265, opened 2026-08-22
docs(postgrest-v3): add the stage 4 implementation plan

grdsdev:
Closes SDK-1559. Stacked on #1264.

Five tasks: inventory the surface, make the new API reachable from `SupabaseClient`, attach the deprecations one file per commit, prove the frozen API is frozen, verify and schedule the deletion.

## Task 2 comes before any deprecation, deliberately

`SupabaseClient.rest` is **internal** and `SupabaseClient.from(_ table: String)` returns `PostgrestQueryBuilder`. Verified against the current source: the only public route to the typed API is `client.schema("public").from(Todo.self)`. Deprecating `from(_:)` while the replacement needs a workaround would be actively hostile, so the overloads §6 calls for land first.

## §6's conclusion survives, but its original argument is dead

The spec's reason for leaving today's implementation untouched rather than reimplementing it over the new core *used to be* aliasing — "value types cannot reproduce the old behavior where two chains branched off one builder mutate the same object". #1240 removed that by making the builders value types.

The reason that still holds is narrower, and the plan says to cite this one in review: the legacy path keeps its hand-rolled retry loop and inconsistent filter escaping, and stages 2–3 fix both in the new core. Reimplementing the deprecated builders over that core would **silently change their behavior during the deprecation window** — exactly what deprecating something is meant to avoid.

The plan states this explicitly so review does not reopen the decision on the strength of the argument that died.

## Turning the freeze into a test

Task 4 asserts the legacy path's known defects still behave as they did — which reads backwards on purpose, and the plan says so in the test file. Same move SDK-1519 made for the macro dependency boundary: an intention nobody checks is not a guarantee.

One subtlety recorded there: the baseline is "legacy as of stage 4", not "legacy as of the spec being written", because SDK-1509 and SDK-1510 deliberately *did* fix filter correctness on the legacy path — they were silent-wrong-rows bugs.

## Scale

113 public declarations across the six legacy files, measured. Hence one file per commit rather than one 113-symbol commit, and `V3_MIGRATION.md` entries grouped by class rather than one per symbol.

Also flagged: `Examples/` is not covered by `swift test`, so migrating it off the deprecated API needs the Xcode schemes built locally before pushing.

Docs only — no code changes.


grdsdev:
@copilot resolve the merge conflicts in this pull request

PR #1275, opened 2026-08-25
fix(postgrest): keep observed stored properties as columns

grdsdev:
Stacked on #1274.

## What

`postgrestStoredProperties()` skipped every binding with a non-nil `accessorBlock`. That filter means to skip computed properties, but observers land in the same block, so a genuinely stored property:

```swift
@Table("todos")
struct Todo {
  @PrimaryKey var id: Int
  var isDone: Bool = false { didSet { print(isDone) } }
}
```

was treated as computed and silently got no column — no `CodingKeys` case, no `columnName(for:)` case, and no field in `Insert` or `Update`.

The filter now inspects the block's contents via a shared `PatternBindingSyntax.postgrestIsStored`:

- accessors that are only `willSet`/`didSet` → stored, keep the column
- an explicit `get`, a `get`+`set` pair, a `_read`/`unsafeAddress` accessor, or a bare code block standing in for an implicit getter → computed, still no column

## Why the diagnostic changed too

`postgrestDiagnoseUnannotatedProperties(macro:in:)` (added in #1274) deliberately mirrors the member filter in `postgrestStoredProperties()`, so it carried the same `accessorBlock == nil` narrowing. Without the same fix, `var isDone = false { didSet { ... } }` — both observed *and* unannotated — stayed silently dropped with no diagnostic. Both call sites now share one helper, which is what keeps them in step.

## How it was tested

Swift Testing, `assertMacro` from MacroTesting:

- `TableMacroTests.keepsObservedStoredProperties` — a `didSet` property and a `willSet` property both get their column, `CodingKeys` case, and `Insert`/`Update` field
- `TableMacroTests.skipsPropertiesThatComputeTheirValue` — regression guard over an explicit getter, a getter with a setter, and a `_read` coroutine
- `SelectionOfMacroTests.keepsObservedStoredProperties` — the observed property lands in `selectString` and in `_columnCheck`
- `DiagnosticsTests.tableRejectsAnObservedPropertyWithNoTypeAnnotation` — an accessor block does not excuse a missing annotation

The three new positive tests were watched failing first; the recorded diff showed the observed properties missing from every generated member and `Insert`/`Update` collapsing to `init()`. `skipsPropertiesThatComputeTheirValue` passed from the start, which is the point — it pins behaviour the fix must not change.

Full suite: 1264 tests pass. `./scripts/format.sh` and `./scripts/spell-check.sh` clean.

## Review notes

The `case .getter` arm covers `var summary: String { htmlURL }`, where SwiftSyntax gives a bare `CodeBlockItemListSyntax` and there is no accessor specifier to inspect. That is the shape the original `accessorBlock == nil` check was really aiming at.

## Risk

Unreleased macros, so no released API changes and no migration-guide entry. The behaviour change is strictly additive for observed properties: a type that previously compiled with a silently missing column now gets that column, which is what the author wrote.

## Known gap, not addressed here

`variable.bindings.first` still means a comma-separated declaration (`var task: String, note: String`) contributes only one column. Tracked separately.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,5 +1,5 @@
 {
-  "originHash" : "e02e24d410edd1a54b0cc6c1b179cf93c6fab7b27de683761e1eb6b27f8e8b8c",
+  "originHash" : "66fc84e1d93fc7d8e6e79e85cb1c6583c07ea159d66325d0329cfeba57a7a06a",
   "pins" : [
     {
       "identity" : "cryptoswift",
@@ -83,6 +83,15 @@
       }
     },
     {
+      "identity" : "swift-macro-testing",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/pointfreeco/swift-macro-testing",
+      "state" : {
+        "revision" : "2e494c632d510715c96a694ff25e2a8d4ac3f64b",
+        "version" : "0.6.5"
+      }
+    },
+    {
       "identity" : "swift-secp256k1",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/21-DOT-DEV/swift-secp256k1.git",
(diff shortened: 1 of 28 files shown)


grdsdev:
Superseded by #1261, which landed this fix directly.

`main` now has `PatternBindingSyntax.isPostgrestStored` in `Sources/PostgrestMacrosPlugin/Support/StoredProperty.swift`, with the same `willSet`/`didSet`-only check this PR proposed, and `TableMacroTests.includesAStoredPropertyWithObservers` covering it.

I based this branch on #1261's tip before that work was added, so the duplication is mine. Nothing here to carry forward.

PR #1276, opened 2026-08-25
fix(postgrest): require one stored property per declaration

grdsdev:
Stacked on #1275.

## What

`postgrestStoredProperties()` reads `bindings.first`, so a comma-separated declaration only ever contributed one column:

```
@Table("todos")
struct Todo {
  @PrimaryKey var id: Int
  var task: String, note: String
      ┬─────────────────────────
      ╰─ 🛑 @Table requires one stored property per declaration; split 'task' and 'note' into separate declarations
}
```

Before this, `note` was silently dropped from `CodingKeys`, `columnName(for:)`, `Insert` and `Update`.

## Why a diagnostic, not iterating the bindings

Iterating `variable.bindings` and emitting a column per binding was the obvious alternative. It doesn't hold up:

- **Marker attributes attach to the declaration, not to a binding.** `@Column("a") var x: Int, y: Int` would name one column twice, and `@PrimaryKey var x: Int, y: Int` would claim two primary keys. Both would have to be rejected anyway, so the ambiguity doesn't go away — it just moves.
- **`var task, note: String` needs the compiler's annotation-propagation rule.** The parser gives one binding with *no* annotation and one with it. I confirmed this against the actual macro before choosing: on `main` + #1274 the shared-annotation form produces

  > 🛑 @Table requires an explicit type annotation on 'task' …

  for a property whose type the author plainly did write. Getting option 1 right means reimplementing that propagation by hand.

One property per declaration removes both problems, and it's the shape this SDK writes anyway.

## Ordering matters

The check runs *ahead* of `postgrestDiagnoseUnannotatedProperties(macro:in:)`, which is what makes `var task, note: String` get the split message instead of the misleading annotation one. `DiagnosticsTests.tableRejectsASharedTypeAnnotation` pins that ordering — it fails if the two checks are ever swapped.

That ordering also makes `bindings.first` correct by construction in both readers rather than a silent narrowing. Both doc comments now say so, so the next person doesn't "fix" it back.

## How it was tested

`Tests/PostgrestMacrosTests/DiagnosticsTests.swift`, Swift Testing with recorded caret art:

- `tableRejectsMoreThanOnePropertyPerDeclaration`
- `tableRejectsASharedTypeAnnotation` — the `var task, note: String` form
- `tableAcceptsAStaticDeclarationThatBindsSeveralNames` — `static let first = "a", second = "b"` maps to no column either way, so splitting it would buy nothing; full expansion snapshot asserts no diagnostic fires
- `selectionOfRejectsMoreThanOnePropertyPerDeclaration` — three names, which also covers the `'a', 'b' and 'c'` joiner

All watched failing first. Full suite: 1268 tests pass. `./scripts/format.sh` and `./scripts/spell-check.sh` clean.

## Risk

Unreleased macros, so no released API changes and no migration-guide entry. This one is a *narrowing* rather than an addition: a declaration that previously compiled with a silently missing column is now an error naming the fix.

With this landed, all three silent-column-drop paths in the macro's property reader are closed — missing annotation (#1274), observers (#1275), and multiple bindings (here).

### DIFF
--- a/Sources/PostgrestMacrosPlugin/SelectionOfMacro.swift
+++ b/Sources/PostgrestMacrosPlugin/SelectionOfMacro.swift
@@ -40,6 +40,9 @@ public struct SelectionOfMacro: ExtensionMacro {
       )
       return []
     }
+    if declaration.postgrestDiagnoseMultipleBindings(macro: "@SelectionOf", in: context) {
+      return []
+    }
     if declaration.postgrestDiagnoseUnannotatedProperties(macro: "@SelectionOf", in: context) {
       return []
     }
--- a/Sources/PostgrestMacrosPlugin/Support/Diagnostics.swift
+++ b/Sources/PostgrestMacrosPlugin/Support/Diagnostics.swift
@@ -56,6 +56,9 @@ extension DeclGroupSyntax {
   /// names the key path and hits the generated `fatalError`. A macro cannot recover the type from
   /// the initializer expression, so the author is asked for an annotation instead.
   ///
+  /// Reads `bindings.first` because ``postgrestDiagnoseMultipleBindings(macro:in:)`` has already
+  /// rejected anything that binds more than one property.
+  ///
   /// Returns `true` if anything was reported, so the caller can stop before emitting an expansion
   /// that leaves the column out.
   func postgrestDiagnoseUnannotatedProperties(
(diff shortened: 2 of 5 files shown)


grdsdev:
Superseded by #1261, and its approach no longer fits.

#1261 fixed the multi-binding bug by taking **option 1** — `postgrestStoredProperties()` now iterates every binding via `flatMap`, and `[PatternBindingSyntax].postgrestType(at:)` resolves a shared annotation forward (`var draft, review: String`) while refusing to for a binding with its own initializer (`var a = 1, b: String`). `TableMacroTests.coversEveryBindingInAMultiBindingDeclaration` covers it.

This PR took option 2 and *rejects* a multi-binding declaration outright, so merging it now would regress a case `main` supports. Closing rather than reworking.

For the record, the objection this PR raised against option 1 was real and #1261 answered it: the shared-annotation form does need the compiler's propagation rule, and `postgrestType(at:)` is where that lives. The remaining attribute ambiguity — `@Column("a") var x: Int, y: Int` naming one column twice, `@PrimaryKey var x: Int, y: Int` claiming two primary keys — is still unguarded on `main`. I have not filed that separately; say the word if it's worth its own issue.

PR #1278, opened 2026-08-25
fix(postgrest): make the primary key optional in Insert, not absent

grdsdev:
Fixes SDK-1606. Stacked on #1274.

## What

`@PrimaryKey` had exactly one effect in `TableMacro.swift`:

```swift
let writable = properties.filter { !$0.isPrimaryKey }
```

Key columns were dropped from `Insert` and `Update`. Right for a single database-generated `id`; wrong everywhere else.

### 1. A compound natural key made the table unwritable

```swift
@Table("user_roles")
struct UserRole {
  @PrimaryKey var userID: UUID
  @PrimaryKey var roleID: UUID
  var grantedAt: Date
}
```

generated `Insert { var grantedAt: Date }` — no way to send either key, so **no row could be created**. A pure join table got an `Insert` with no fields and an empty `init()`.

### 2. It broke `upsert`, for every table

`upsert(_ values: R.Insert, onConflict: String? = nil)` sends no `on_conflict` by default. Per [PostgREST's docs](https://docs.postgrest.org/en/stable/references/api/tables_views.html#upsert), upsert then operates on the primary key columns and "you must specify all of them" in the body. `Insert` never carried them — so the merge half of an upsert was unreachable and every call just inserted another row.

Nothing about either failure was loud: no compile error, no diagnostic.

## Why option 3

The marker conflated two facts that only coincide for a serial `id`:

1. this column identifies the row — what the marker is named for
2. the database generates it, so don't send it — what the filter implemented

(2) is already what `@Default` means — so `@Default` is what expresses it, and `@PrimaryKey` no longer does.

`Insert` now carries every column, and a column is optional exactly when the database can fill it in: it is nullable, or it has a default. That is `postgres-meta`'s rule verbatim — `is_nullable || is_identity || default_value !== null` — which is what types supabase-js from the same column metadata, and it never consults the primary key either.

A generated key is spelled with both markers. The plugin already read them independently, so this needed no plugin change:

```swift
@PrimaryKey @Default var id: Int      // generated — omit it on insert
@PrimaryKey var code: String          // natural — required on insert
```

```swift
Todo.Insert(task: "buy milk")            // id omitted, the database fills it (@Default)
Todo.Insert(id: 7, task: "buy milk")     // now possible
UserRole.Insert(userID: 1, roleID: 2)    // now possible
UserRole.Insert(userID: 1)               // compile error — half a key
UserRole.Insert()                        // compile error
```

> **Revised after review.** The first revision used `isOptional || hasDefault || isPrimaryKey`, making *every* key optional. [Review feedback](https://github.com/supabase/supabase-swift/pull/1278#pullrequestreview-5018711244) established that this diverges from supabase-js in the one place type safety matters, and that the "costs existing users nothing" argument for it does not apply to an API v2.55.1 never shipped. Changed in `cc2a3bb`.

`Update` is deliberately unchanged — it targets rows by key rather than changing it, so the key stays out. **Worth a reviewer opinion:** the same argument that motivates this PR also applies to *changing* a natural key, which `Update` still can't express. Kept out of scope here; happy to follow up.

## What this does not fix

This makes a correct `upsert` possible; it does not make an incorrect one impossible:

```swift
try client.from(Todo.self).upsert(Todo.Insert(task: "buy milk"))  // still silently just inserts
```

With no `onConflict:` and no key in the body, PostgREST has nothing to merge on. supabase-js has the same hole, so this is parity-neutral — but the typed layer exists to close exactly this kind of hole. A typed `onConflict: [PartialKeyPath<R>]`, or deriving `on_conflict` from the `@PrimaryKey` columns, would close it. Filed as [SDK-1616](https://linear.app/supabase/issue/SDK-1616/postgrest-typed-upsert-compiles-with-no-conflict-target-and-silently) — which turned up a prerequisite worth knowing: nothing exposes the primary key at runtime, since `@PrimaryKey` is consumed entirely at expansion time, so `@Table` would need to emit the key columns first.

## How it was tested

New, each watched failing first:

- `TableMacroTests.insertCarriesACompoundPrimaryKey` — before the fix the expansion was `init(grantedAt: Date)`
- `TableMacroTests.insertOnATableThatIsNothingButACompoundKey` — the pure join table
- `TableIntegrationTests.insertCarriesAPrimaryKeyTheClientSupplies` — `"id":7` reaches the wire
- `TableIntegrationTests.insertCarriesAWholeCompoundPrimaryKey` — `"user_id":1` and `"role_id":2` both reach the wire
- `TableIntegrationTests.aTypedUpsertCarriesTheConflictKey` — the upsert body now contains its own conflict target; there was no typed-upsert test at all before

The integration tests are the strongest evidence: against the pre-fix macro they don't *compile* — `extra argument 'id' in call` and `extra arguments at positions #1, #2 in call`.

> One note for anyone re-checking this locally: SwiftPM does **not** recompile a module when only the macro plugin changed, so reverting `TableMacro.swift` and re-running appears to pass against a stale expansion. Touch the test file to force it.

`insertExcludesThePrimaryKeyAndUsesColumnNames` is renamed to `insertOmitsANilPrimaryKeyAndUsesColumnNames` — its assertions still hold, but the key is now omitted-when-nil rather than excluded, and the old name claimed otherwise.

Existing expansion snapshots gained the key as an `Insert` field — required or optional depending on `@Default`. All re-recorded, not hand-edited, including the one inside `DiagnosticsTests` that also captures an expansion.

Full suite: 1269 tests pass. `./scripts/format.sh`, `./scripts/spell-check.sh`, and `./scripts/test-docs.sh` all clean.

## Docs corrected

Four comments asserted a behaviour this PR removes:

- `@PrimaryKey`: "Marks a property as the relation's primary key, excluding it from the generated `Insert`." Then, after the first revision, "making it optional in the generated `Insert`" — which would have shipped as a lie once the rule changed. It now says what the marker does and does not do, with both spellings side by side.
- `insert(_:)`: "Primary keys and defaulted columns are absent or optional there."
- `PostgrestWritableRelation.Insert`: "primary keys excluded, defaulted and nullable columns optional." Wrong twice over — keys are included, and their optionality follows `@Default`.
- The rationale above it: "Defaulting them to `Self` would mean sending the primary key on insert." Sending it is now intended, so the reasoning read backwards.

`@Default` also gains a line noting it is the only marker that makes a column optional, and the `@Table` example carries `@PrimaryKey @Default var id: Int` so the common case is what a reader copies.

## Not a breaking change

No `!` and no `V3_MIGRATION.md` entry: `PostgrestMacros` does not exist in v2.55.1, so this changes nothing a released version ever exposed. That matches how the rest of the macro stack was marked, against `refactor(postgrest)!` for the builders that did ship.

### DIFF
--- a/Sources/PostgREST/Query/PostgrestTypedMutation.swift
+++ b/Sources/PostgREST/Query/PostgrestTypedMutation.swift
@@ -57,7 +57,8 @@ extension PostgrestTypedSource where R: PostgrestWritableRelation {
   /// Inserts a row.
   ///
   /// - Parameter values: The row to insert, in the relation's `Insert` shape. Primary keys and
-  ///   defaulted columns are absent or optional there.
+  ///   defaulted columns are optional there, so either can be left out and filled in by the
+  ///   database.
   /// - Returns: A ``PostgrestTypedMutation`` to execute, or to request rows back from.
   /// - Throws: An encoding error if `values` cannot be serialized.
   public func insert(_ values: R.Insert) throws -> PostgrestTypedMutation<R> {
--- a/Sources/PostgREST/Relations/PostgrestRelation.swift
+++ b/Sources/PostgREST/Relations/PostgrestRelation.swift
@@ -42,11 +42,13 @@ public protocol PostgrestRelation: PostgrestSelection where Source == Self {
 
 /// A relation the database accepts writes for: a table, or a view Postgres reports as updatable.
 ///
-/// `Insert` and `Update` have no defaults on purpose. Defaulting them to `Self` would mean sending
-/// the primary key on insert and requiring every column on update.
+/// `Insert` and `Update` have no defaults on purpose. Defaulting them to `Self` would mean
+/// requiring every column on both — including the ones the database fills in on insert, and the
+/// ones an update is not touching.
 public protocol PostgrestWritableRelation: PostgrestRelation {
-  /// The shape accepted by an insert: primary keys excluded, defaulted and nullable columns
-  /// optional.
+  /// The shape accepted by an insert: every column, optional exactly where the database can fill
+  /// it in — a nullable column, or one with a default. A primary key is included, and required
+  /// unless it is also defaulted.
   associatedtype Insert: Encodable & Sendable
 
   /// The shape accepted by an update: every column optional.
--- a/Sources/PostgrestMacros/Macros.swift
+++ b/Sources/PostgrestMacros/Macros.swift
@@ -56,7 +56,14 @@ public macro Column(_ name: String) =
     module: "PostgrestMacrosPlugin", type: "MarkerMacro"
   )
 
-/// Marks a property as the relation's primary key, excluding it from the generated `Insert`.
+/// Marks a property as part of the relation's primary key, making it optional in the generated
+/// `Insert`.
+///
+/// Leaving it `nil` omits the column, so a database-generated key fills itself in. Supplying it
+/// sends it, which is what a compound natural key needs — nothing in the database generates those.
+///
+/// Apply it to more than one property for a compound key. The key is left out of `Update`, which
+/// targets rows by key rather than changing it.
 @attached(peer)
 public macro PrimaryKey() =
   #externalMacro(
--- a/Sources/PostgrestMacros/Macros.swift
+++ b/Sources/PostgrestMacros/Macros.swift
@@ -14,7 +14,7 @@
 /// ```swift
 /// @Table("todos")
 /// struct Todo {
-///   @PrimaryKey var id: Int
+///   @PrimaryKey @Default var id: Int
 ///   var task: String
 ///   @Default var isDone: Bool
 ///   @Column("due_at") var dueDate: Date?
@@ -56,14 +56,23 @@ public macro Column(_ name: String) =
     module: "PostgrestMacrosPlugin", type: "MarkerMacro"
   )
 
-/// Marks a property as part of the relation's primary key, making it optional in the generated
-/// `Insert`.
+/// Marks a property as part of the relation's primary key.
 ///
-/// Leaving it `nil` omits the column, so a database-generated key fills itself in. Supplying it
-/// sends it, which is what a compound natural key needs — nothing in the database generates those.
+/// This does **not** make the column optional in `Insert`. Being the key says nothing about
+/// whether the database can supply the value — a `SERIAL` or identity column can, and a natural
+/// key like a country code cannot. Add ``Default()`` for the ones it can:
 ///
-/// Apply it to more than one property for a compound key. The key is left out of `Update`, which
-/// targets rows by key rather than changing it.
+/// ```swift
+/// @PrimaryKey @Default var id: Int      // generated — omit it on insert
+/// @PrimaryKey var code: String          // natural — required on insert
+/// ```
+///
+/// Apply it to more than one property for a compound key. Each half is then required unless it is
+/// also ``Default()``, so an incomplete key is a compile error rather than a request PostgREST
+/// rejects. This matches how `postgres-meta` types supabase-js: it makes a column optional from
+/// `is_nullable || is_identity || default_value !== null`, and never consults the primary key.
+///
+/// The key is left out of `Update`, which targets rows by key rather than changing it.
 @attached(peer)
 public macro PrimaryKey() =
   #externalMacro(
@@ -72,7 +81,9 @@ public macro PrimaryKey() =
 
 /// Marks a property as having a database default, making it optional in the generated `Insert`.
 ///
-/// Leaving it `nil` omits the column from the request body, so the database fills it in.
+/// Leaving it `nil` omits the column from the request body, so the database fills it in. This is
+/// the only marker that makes a column optional — including for a primary key, which needs
+/// `@PrimaryKey @Default` when the database generates it.
 @attached(peer)
 public macro Default() =
   #externalMacro(
--- a/Sources/PostgrestMacrosPlugin/Support/StoredProperty.swift
+++ b/Sources/PostgrestMacrosPlugin/Support/StoredProperty.swift
@@ -23,6 +23,19 @@ struct StoredProperty {
   var optionalType: String { isOptional ? type : "\(type)?" }
 }
 
+/// Whether a written type is spelled as an `Optional`.
+///
+/// Both spellings have to be recognized everywhere optionality is decided: `Int?` and
+/// `Optional<Int>` are the same type, and a generated schema can emit either. Recognizing one and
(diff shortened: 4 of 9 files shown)


mandarini (approved):

grdsdev (commented):
## Review: checked against the PostgREST docs and postgres-meta

I verified the premise against the PostgREST docs and cross-checked the generated shapes against `postgres-meta`, which is the ground truth for supabase-js types. The macro suite passes at the tip of the stack: **43 tests, exit 0.**

### What the sources confirm

**PostgREST — the upsert argument is correct.** The docs say, verbatim: "By default, upsert operates based on the primary key columns, so you must specify all of them." Dropping the key from `Insert` did make the merge half of every `upsert` unreachable. That part of the PR body holds up.

**postgres-meta** — [`src/server/templates/typescript.ts`](https://github.com/supabase/postgres-meta/blob/master/src/server/templates/typescript.ts) lines 565–601:

```ts
Insert: { ... is_optional: column.is_nullable || column.is_identity || column.default_value !== null }
Update: { ... is_optional: true }
// both: if (column.identity_generation === 'ALWAYS') return `"name"?: never`
```

Three facts follow: the primary key is **never** consulted; `Update` makes every column optional; and in `Insert` a NOT NULL primary key with no default is **required**.

---

### 1. The `Insert` rule is looser than supabase-js, and the reason given for it no longer holds

This PR uses `isOptional || hasDefault || isPrimaryKey`. postgres-meta's rule is the same expression **without** `isPrimaryKey` — `@Default` already carries exactly what `is_identity || default_value !== null` means.

The cost shows up in this PR's own headline table:

```swift
IntegrationUserRole.Insert(userID: 1)   // compiles. body omits role_id -> runtime 400
IntegrationUserRole.Insert()            // also compiles. body is {}
```

Both fail type-check in supabase-js. The `init(userID: UUID? = nil, roleID: UUID? = nil)` in `insertOnATableThatIsNothingButACompoundKey` still accepts zero arguments, so the "`Insert` with no fields and an empty `init()`" complaint in the description is only half retired — the fields are there now, but the empty call still compiles.

The PR justifies optional-over-required with "costs existing users nothing." But the same PR argues, correctly, that `PostgrestMacros` does not exist in v2.55.1 — so there are no existing users. That removes the deciding factor for option 3.

**Suggestion:** drop `|| $0.isPrimaryKey` and spell a generated key `@PrimaryKey @Default var id: Int`. The two markers already compose with no plugin change (`attribute("PrimaryKey")` and `attribute("Default")` are read independently). That buys exact JS parity, forces both halves of a compound key at compile time, and is what a schema generator would emit from the same column metadata postgres-meta reads.

If an extra `@Default` on every `id` reads as too noisy, a fallback that keeps the zero-annotation-cost property: make the key required when **more than one** property is marked, optional when exactly one. It encodes "one surrogate key is generated, a compound key is natural." It gets a single *natural* key wrong (`@PrimaryKey var code: String` on a country table), so I'd take the first option.

### 2. This makes a correct upsert possible, not an incorrect one impossible

`aTypedUpsertCarriesTheConflictKey` proves the good call now works. The bad call still compiles:

```swift
try client.from(Todo.self).upsert(Todo.Insert(task: "buy milk"))  // still silently just inserts
```

supabase-js has the same hole, so this is parity-neutral — but the typed layer exists to close exactly this kind of hole. A typed `onConflict: [PartialKeyPath<R>]`, or deriving `on_conflict` from the `@PrimaryKey` columns, would close it. Worth naming as a follow-up in the description rather than leaving the impression `upsert` is fully fixed.

### 3. Two stale doc comments this PR missed

The description says two comments asserted the old behaviour and were corrected. There are two more, in `Sources/PostgREST/Relations/PostgrestRelation.swift`:

- **line 48** — `/// The shape accepted by an insert: primary keys excluded, defaulted and nullable columns optional.` Now false.
- **line 46** — `/// Defaulting them to Self would mean sending the primary key on insert and requiring every column on update.` Sending the key on insert is now the intended behaviour, so the rationale reads backwards.

That is the associated-type doc a hand-written conformance reads first. Both belong in this PR.

### 4. New footgun: `GENERATED ALWAYS AS IDENTITY`

postgres-meta emits `?: never` for those columns in **both** `Insert` and `Update`, so supabase-js rejects them at compile time. Before this PR the key's absence from `Insert` protected the Swift shape by accident. Now it is accepted and Postgres rejects the write at runtime (`428C9`). A `@Generated` marker that excludes the column from both shapes would restore parity; at minimum a doc line on `@PrimaryKey`.

Filed as [SDK-1609](https://linear.app/supabase/issue/SDK-1609/postgrest-table-insert-and-update-accept-a-generated-always-identity).

### 5. Minor

- `upsert(_ values: R.Insert)` documents `values` as "The rows to upsert." It takes one row. (Separate scope: the typed API has no bulk insert at all, while supabase-js accepts an array — but the wording should be singular today.)
- `@PrimaryKey var id: Int?` is accepted silently. A primary key is never nullable in Postgres. Cheap diagnostic.

Both filed as [SDK-1611](https://linear.app/supabase/issue/SDK-1611/postgrest-typed-api-doc-and-diagnostic-polish-from-the-12781280-review).

The bulk-insert gap noted in the first bullet is filed separately as [SDK-1612](https://linear.app/supabase/issue/SDK-1612/postgrest-typed-insertupsert-accept-only-one-row-no-bulk-insert). Researching it turned up a hard constraint worth knowing here: PostgREST fills a key that is missing from one object in a bulk payload with NULL, not the column default, unless `Prefer: missing=default` is sent. Since the generated `Insert` omits nil optionals — the whole point of `@Default` — a typed bulk insert would silently NULL out defaulted columns. Swift sends no `missing=` header anywhere today, which is [SDK-1613](https://linear.app/supabase/issue/SDK-1613/postgrest-no-defaulttonull-prefer-missingdefault-on-insert-and-upsert), a prerequisite for the bulk work.

---

### Verdict

Approve the direction, change the rule. The bug is real, the docs back the upsert claim, and the tests were genuinely watched failing. But `|| isPrimaryKey` diverges from supabase-js in the one place type safety matters, and the backward-compatibility argument for it does not apply to an unreleased API. Findings 1, 2 and 3 are the ones I'd want resolved before merge; 4 and 5 are follow-ups.

**Follow-ups filed:** [SDK-1609](https://linear.app/supabase/issue/SDK-1609/postgrest-table-insert-and-update-accept-a-generated-always-identity) · [SDK-1611](https://linear.app/supabase/issue/SDK-1611/postgrest-typed-api-doc-and-diagnostic-polish-from-the-12781280-review) · [SDK-1612](https://linear.app/supabase/issue/SDK-1612/postgrest-typed-insertupsert-accept-only-one-row-no-bulk-insert) · [SDK-1613](https://linear.app/supabase/issue/SDK-1613/postgrest-no-defaulttonull-prefer-missingdefault-on-insert-and-upsert)

PR #1280, opened 2026-08-25
fix(postgrest): let Update change a primary key column

grdsdev:
Fixes SDK-1607. Stacked on #1278.

## What

`Update` dropped primary-key columns along with `Insert`, so a key could never be changed through the typed API:

```swift
try client.from(UserRole.self)
  .update(UserRole.Update(roleID: 3))   // extra argument 'roleID' in call
  .eq(\.userID, 1)
```

Renaming a natural key — the `slug` in a table keyed on `(tenant_id, slug)` — meant dropping to the legacy builder.

## Why the exclusion bought nothing

`update(_:)` returns a `PostgrestTypedMutation` the caller filters separately, so targeting and payload are independent concerns. Its own doc already warns that with no filter it updates every row. Holding the key back never stopped a write from rewriting identity — it only removed the ability to do so deliberately.

Every `Update` field is already optional and defaults to `nil`, so this is purely additive: a caller who doesn't name the key is unaffected, and the column stays out of the request body.

## The counter-argument, for the record

Excluding the key gives the typed API a property: a `.update()` can never rewrite a row's identity. That is coherent, and changing a primary key is often a smell — delete and re-insert is usually better.

Rejected as the deciding factor because the protection is narrow (the same call with no filter can already overwrite every *other* column) and it costs a legitimate operation with no typed alternative. Flagging it explicitly since this is the one judgment call in the PR — if you'd rather keep the guard, this is the commit to drop.

Deliberately split from #1278 rather than folded in: that one fixed "this table cannot be written to at all" plus a silently degraded `upsert`. This is a missing capability with a workaround, and it deserved its own review rather than riding along on a case that wasn't in doubt.

## Result

The `writable` filter is gone.

> **Revised after review.** That left `isPrimaryKey` with *no* behavioural use at all. This PR removed the `Update` filter, and [#1278's revision](https://github.com/supabase/supabase-swift/pull/1278) removed `|| isPrimaryKey` from the `Insert` rule, so after rebasing, writing `@PrimaryKey` or omitting it expanded byte-for-byte the same — the marker was inert. [Review finding 1](https://github.com/supabase/supabase-swift/pull/1280#pullrequestreview-5018712411) called that out, and it was not filed as a follow-up because it is a decision for these two PRs.
>
> Resolved by giving the marker the one job only it can do: `PostgrestRelation.primaryKeyColumns`, the key's column names in declaration order.
>
> ```swift
> static let primaryKeyColumns: [String] = ["user_id", "role_id"]
> ```
>
> Emitted only when a key is declared, so a keyless relation falls through to a protocol default of `[]` and its expansion is unchanged. The requirement carries that default rather than being mandatory, so hand-written conformances keep compiling. This is also the prerequisite [SDK-1616](https://linear.app/supabase/issue/SDK-1616/postgrest-typed-upsert-compiles-with-no-conflict-target-and-silently) needs to derive an upsert's `on_conflict` — previously unbuildable, because the key never survived expansion.

## How it was tested

Both watched failing first:

- `TableIntegrationTests.updateCanChangeAKeyColumn` — `{"role_id":3}` reaches the wire, filtered by `.eq(\.userID, 1)`. Against the pre-fix macro this doesn't compile: `extra argument 'roleID' in call`.
- `TableIntegrationTests.updateStillOmitsAKeyItIsNotChanging` — `Todo.Update(task:)` still sends `{"task":"buy oat milk"}` and nothing else, so the addition costs existing callers nothing.

Seven existing expansion snapshots gained the key as an optional `Update` field. All re-recorded, not hand-edited.

Full suite: 1271 tests pass. `./scripts/format.sh`, `./scripts/spell-check.sh`, and `./scripts/test-docs.sh` all clean.

## Docs corrected

The `@PrimaryKey` doc added in #1278 said the key "is left out of `Update`, which targets rows by key rather than changing it." That is now false and is replaced.

## Not a breaking change

No `!` and no `V3_MIGRATION.md` entry: `PostgrestMacros` does not exist in v2.55.1, so this changes nothing a released version ever exposed.

### DIFF
--- a/Sources/PostgREST/Relations/PostgrestRelation.swift
+++ b/Sources/PostgREST/Relations/PostgrestRelation.swift
@@ -38,6 +38,21 @@ public protocol PostgrestRelation: PostgrestSelection where Source == Self {
   /// - Parameter keyPath: A key path to one of this type's stored properties.
   /// - Returns: The column name PostgREST expects in a query string.
   static func columnName<V>(for keyPath: KeyPath<Self, V>) -> String
+
+  /// The columns making up the relation's primary key, in declaration order.
+  ///
+  /// Empty when the relation declares no key. A caller deriving a conflict target for an upsert
+  /// has to treat that as "cannot derive" rather than "conflicts on nothing" — an empty
+  /// `on_conflict` is not the same request.
+  static var primaryKeyColumns: [String] { get }
+}
+
+extension PostgrestRelation {
+  /// No declared key.
+  ///
+  /// `@Table` overrides this whenever a property is marked `@PrimaryKey`; a hand-written
+  /// conformance opts in by declaring it.
+  public static var primaryKeyColumns: [String] { [] }
 }
 
 /// A relation the database accepts writes for: a table, or a view Postgres reports as updatable.
--- a/Sources/PostgrestMacros/Macros.swift
+++ b/Sources/PostgrestMacros/Macros.swift
@@ -72,7 +72,8 @@ public macro Column(_ name: String) =
 /// rejects. This matches how `postgres-meta` types supabase-js: it makes a column optional from
 /// `is_nullable || is_identity || default_value !== null`, and never consults the primary key.
 ///
-/// The key is left out of `Update`, which targets rows by key rather than changing it.
+/// `Update` carries the key like any other column, all optional, so a natural key can be renamed.
+/// Which rows a write touches is decided by the filters on the mutation, not by this marker.
 @attached(peer)
 public macro PrimaryKey() =
   #externalMacro(
--- a/Sources/PostgrestMacros/Macros.swift
+++ b/Sources/PostgrestMacros/Macros.swift
@@ -37,7 +37,7 @@
   extension,
   conformances: Decodable, Sendable, PostgrestRelation, PostgrestWritableRelation,
   names: named(relationName), named(schema), named(selectString), named(columnName(for:)),
-  named(CodingKeys), named(Insert), named(Update)
+  named(primaryKeyColumns), named(CodingKeys), named(Insert), named(Update)
 )
 public macro Table(
   _ name: String,
@@ -74,6 +74,10 @@ public macro Column(_ name: String) =
 ///
 /// `Update` carries the key like any other column, all optional, so a natural key can be renamed.
 /// Which rows a write touches is decided by the filters on the mutation, not by this marker.
+///
+/// What the marker does produce is ``PostgREST/PostgrestRelation/primaryKeyColumns``, the column
+/// names in declaration order. That is what lets a caller — or a future `upsert` — name a conflict
+/// target without repeating the key as a string.
 @attached(peer)
 public macro PrimaryKey() =
   #externalMacro(
--- a/Sources/PostgrestMacrosPlugin/TableMacro.swift
+++ b/Sources/PostgrestMacrosPlugin/TableMacro.swift
@@ -81,6 +81,15 @@ public struct TableMacro: ExtensionMacro {
       "  \(access)static let selectString = \"*\"",
       columnNameFunction(access: access, type: type.trimmedDescription, properties: properties),
     ]
+    // The one thing `@PrimaryKey` uniquely does. Without this the marker is inert: after the key
+    // stopped gating `Insert` optionality and stopped being filtered out of `Update`, writing it or
+    // omitting it expanded byte-for-byte the same. Emitted only when a key is declared, so a
+    // keyless relation falls back to the protocol's empty default.
+    let keyColumns = properties.filter(\.isPrimaryKey).map(\.columnName)
+    if !keyColumns.isEmpty {
+      let list = keyColumns.map { "\"\($0)\"" }.joined(separator: ", ")
+      body.append("  \(access)static let primaryKeyColumns: [String] = [\(list)]")
+    }
     if let codingKeys = codingKeys(for: properties) {
       body.append(codingKeys)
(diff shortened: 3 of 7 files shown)


grdsdev (commented):
## Review: this one matches postgres-meta exactly

I cross-checked against `postgres-meta`, which is the ground truth for the supabase-js generated types. The macro suite passes at the tip of the stack: **43 tests, exit 0.**

[`src/server/templates/typescript.ts`](https://github.com/supabase/postgres-meta/blob/master/src/server/templates/typescript.ts) lines 585–601:

```ts
Update: {
  ${columnsByTableId[table.id].map((column) => {
    if (column.identity_generation === 'ALWAYS') { return `"name"?: never` }
    return generateColumnTsDefinition(schema, { ..., is_optional: true }, ...)
  })}
}
```

Every column optional, key included, and the primary key is **never** consulted anywhere in the generator. That is precisely what this PR produces. The "why the exclusion bought nothing" argument is also the right one: targeting is the filter's job, and postgres-meta encodes the same separation by not modelling keys at all.

The doc comment on `PostgrestWritableRelation.Update` — "the shape accepted by an update: every column optional" — was already slightly wrong before this PR, and this PR makes it true.

---

### 1. `@PrimaryKey` and `@Default` now generate identical code

After this PR, `isPrimaryKey` has exactly one use in the whole plugin:

```
Sources/PostgrestMacrosPlugin/TableMacro.swift:102
  ($0, $0.isOptional || $0.hasDefault || $0.isPrimaryKey ? $0.optionalType : $0.type)
```

It sits in an `||` next to `hasDefault`. So `@PrimaryKey var id: Int` and `@Default var id: Int` expand byte-for-byte the same. Two markers, one behaviour, and nothing in the output distinguishes them.

The description frames this as the good outcome — "`isPrimaryKey` is now used for exactly one thing" — but that one thing is already fully covered by another marker. Two ways out:

- Differentiate them: drop `|| isPrimaryKey` so `Insert` matches postgres-meta's rule and a generated key is spelled `@PrimaryKey @Default`. (Raised in more detail on #1278, since that is where the rule lands.)
- Give `@PrimaryKey` a job it uniquely can do: derive `on_conflict` from the key columns, or a typed `onConflict: [PartialKeyPath<R>]`. `upsert` still takes `onConflict: String?` while the macro already knows the key columns.

Either way, `@PrimaryKey` shouldn't ship as a decorative marker.

Not filed — this is a decision for this PR and #1278, not a follow-up.

### 2. `Update` still cannot set a nullable column to NULL

Pre-existing rather than introduced here, but this is the PR that settles `Update`'s shape, so it's worth deciding now rather than changing `Update` a third time later.

`optionalType` is `isOptional ? type : "\(type)?"` — no double wrap. So for `@Column("due_at") var dueDate: Date?` the `Update` field is `Date?`, and `Update(dueDate: nil)` means "do not change this" and is omitted from the body. There is no way to express "clear this column."

supabase-js can: the TS type is `string | null` and optional, so `{ due_at: null }` is expressible and `JSON.stringify` keeps the explicit null. Closing the gap needs either a double optional or a `.null` sentinel in the write shapes — both change `Update` again, which is why it's better decided against this PR than after it.

Filed as [SDK-1610](https://linear.app/supabase/issue/SDK-1610/postgrest-typed-update-cannot-set-a-nullable-column-to-null), with a third option that leaves `Update`'s shape alone (`.update(...).clearing(\.dueDate)`) so it does not have to block this PR.

### 3. One doc fix that belongs on #1278, noted here for completeness

`Sources/PostgREST/Relations/PostgrestRelation.swift:46` and `:48` still describe the old `Insert` behaviour ("primary keys excluded"; "would mean sending the primary key on insert"). Both went stale in #1278, not here — I've left the detail on that PR.

---

### Verdict

Approve. This matches postgres-meta's `Update` rule exactly, the reasoning is right, the counter-argument is stated fairly rather than buried, and splitting it from #1278 was the correct call — the two changes really do carry different risk. Finding 1 is the one worth acting on, and it may resolve itself via the `Insert` rule change discussed on #1278. Finding 2 is a genuine JS-parity gap in the same surface, worth a decision now.

**Follow-up filed:** [SDK-1610](https://linear.app/supabase/issue/SDK-1610/postgrest-typed-update-cannot-set-a-nullable-column-to-null). The full set from reviewing both PRs: [SDK-1609](https://linear.app/supabase/issue/SDK-1609/postgrest-table-insert-and-update-accept-a-generated-always-identity) · [SDK-1610](https://linear.app/supabase/issue/SDK-1610/postgrest-typed-update-cannot-set-a-nullable-column-to-null) · [SDK-1611](https://linear.app/supabase/issue/SDK-1611/postgrest-typed-api-doc-and-diagnostic-polish-from-the-12781280-review) · [SDK-1612](https://linear.app/supabase/issue/SDK-1612/postgrest-typed-insertupsert-accept-only-one-row-no-bulk-insert) · [SDK-1613](https://linear.app/supabase/issue/SDK-1613/postgrest-no-defaulttonull-prefer-missingdefault-on-insert-and-upsert)

QuintinWillison (approved):
If I understand correctly, the silent 'drop' of the primary key (via `isPrimaryKey` filter) could be described as "client-side opinion/gate" on something that neither PostgREST or PostgreSQL were even declaring should be disallowed or otherwise blocked.

QuintinWillison:
Oh, and then I just read https://github.com/supabase/supabase-swift/pull/1278, which seemed to be where the drop appeared... so a little confused in aggregate. I'm assuming this is a bit of a journey. 🤔

PR #1288, opened 2026-08-27
feat(postgrest): add column expressions and the filter tree

grdsdev:
Stack 1/8 for SDK-1624. Base: `main`.

Replaces the key-path filter surface with a generated `Columns` namespace, so casts, JSON paths, aggregates and embedded relations are all expressible and compiler-checked. This PR is the vocabulary; nothing consumes it yet.

**`PostgrestColumnExpression`** is anything valid in a `select` list, with two independent refinements — `PostgrestFilterableExpression` (left of an operator) and `PostgrestOrderableExpression` (an `order` key) — because live PostgREST accepts a different expression set in each position. `Root` stops an expression from one relation being used against another; `Value` makes `.eq("seven")` on an `Int` column a compile error.

**`PostgrestFilter`** is the tree: `&&`, `||`, `!`, rendered to query items. Rendering is position-dependent and was corrected against PostgREST 16.1:

- a top-level operand goes unescaped, the same operand inside `or=(…)` must be escaped;
- a negated leaf in a group is `id.not.eq.2` — `not.id.eq.2` is a PGRST100 parse error.

Comparison, null and `raw` operators come with it.

`where(_:)` arrives in 4/8, so the doc examples here reference it ahead of time.

### DIFF
--- a/Sources/PostgREST/Columns/PostgrestColumnExpression.swift
+++ b/Sources/PostgREST/Columns/PostgrestColumnExpression.swift
@@ -40,27 +40,45 @@ public protocol PostgrestFilterableExpression<Root, Value>: PostgrestColumnExpre
 /// embedded columns do not — PostgREST rejects all three in `order`.
 public protocol PostgrestOrderableExpression<Root, Value>: PostgrestColumnExpression {}
 
-/// A stored column, reached through a relation's generated `Columns` namespace.
-public struct PostgrestColumn<Root: PostgrestRelation, Value>: PostgrestFilterableExpression,
-  PostgrestOrderableExpression
-{
-  public let postgrestExpression: String
-
-  /// - Parameter name: The database column name.
-  public init(_ name: String) {
-    self.postgrestExpression = name
-  }
-}
-
-/// A stored column the database allows to be `NULL`.
+/// A filterable expression whose value the database allows to be `NULL`.
 ///
-/// `Value` is the **wrapped** type: `var dueDate: Date?` generates
-/// `PostgrestNullableColumn<Todo, Date>`. So every operator takes a non-optional operand — use
-/// ``PostgrestNullableColumn/isNull()`` to test for `NULL` — while an update accepts `nil` here
-/// to clear the column.
-public struct PostgrestNullableColumn<
+/// Carries ``PostgrestNullableExpression/isNull()``, which is meaningless on a `NOT NULL` column:
+/// it can never match, and a filter that silently returns nothing is worse than one that does not
+/// compile.
+///
+/// This is a protocol rather than a member of one concrete column type so that a future nullable
+/// expression — a JSON path into a nullable column, an outer-joined embed — gets `isNull()` by
+/// conforming, instead of the operator having to be redeclared on it.
+public protocol PostgrestNullableExpression<Root, Value>: PostgrestFilterableExpression {}
+
+// MARK: - Nullability
+
+/// Whether a stored column's database type admits `NULL`.
+///
+/// A phantom type on ``PostgrestStoredColumn``, never a value and never on the wire: it exists so
+/// one struct serves both column kinds while `isNull()` and an update's `nil` assignment stay
+/// available on exactly the nullable one.
+public protocol PostgrestNullability: Sendable {}
+
+/// A `NOT NULL` column.
+public enum PostgrestNotNull: PostgrestNullability {}
+
+/// A column the database allows to be `NULL`.
+public enum PostgrestNullable: PostgrestNullability {}
+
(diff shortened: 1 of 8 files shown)


Tr00d (approved):

spydon (changes requested):
Three findings worth fixing:

**1. `isNull()` is bound to the concrete struct rather than a nullability refinement.** It sits on `PostgrestNullableColumn` (`PostgrestFilter+Comparison.swift:86`), but #1293 adds `PostgrestJSONPath`, which conforms to `PostgrestFilterableExpression` and is very often null (a missing key yields SQL `NULL`). There's no way to spell `data->>k=is.null` on the typed surface, so the only route is `raw("is.null")`, which is the escape hatch this vocabulary exists to avoid. A `PostgrestNullableExpression` refinement carrying `isNull()` is cheaper to introduce now than after four dependent PRs are stacked on top. Related: `PostgrestColumn` and `PostgrestNullableColumn` are identical apart from the name, so a single struct with a nullability phantom parameter would remove the duplication and give `isNull()` a natural home, the same mechanism `isTrue()`/`isFalse()` already use.

**2. The documented rule for `PostgrestFilter.raw(_:_:)` is too weak.** `PostgrestFilter.swift:62` says "Combine this with `&&`, not `||`", but whether a node goes through `group(_:)` depends on whether anything *above* it is an `or` or a `not`, not on its immediate junction. Both of these render the raw node in group position and hit PGRST100:

- `(a && .raw("cost::text", "eq.10")) || b` (the `and` child is rendered by `group`, recursing into the raw case at `:161`)
- `!(a && .raw("cost::text", "eq.10"))` (top-level `not.and` also maps children through `group`)

The accurate rule is that a raw node whose text isn't valid in the group grammar must not appear anywhere beneath an `or` or a `not`. Worth stating that way given the failure mode is a 400 or a silent zero-row answer rather than a compile error. `PostgrestFilterableExpression.raw(_:)` (`PostgrestFilter+Comparison.swift:62`) has the same hazard for its operand and documents nothing about it.

**3. Operator strings are re-hardcoded.** `PostgrestFilter+Comparison.swift` writes `"eq"`, `"neq"`, `"gt"`, `"gte"`, `"lt"`, `"lte"`, `"isdistinct"` and `"is"` as literals, while `PostgrestOperator` (`Legacy/PostgrestFilterBuilder.swift:17`) already defines every one of them in the same module. Two copies of the operator vocabulary is small now and an annoyance once #1289 adds the other twenty.

grdsdev:
All three addressed in be74fa28.

**1. Collapsed the two column structs.** One `PostgrestStoredColumn<Root, Value, Nullability>` with a nullability phantom; `PostgrestColumn` and `PostgrestNullableColumn` stay as typealiases that fix it, so call sites and the shape the generator will emit are unchanged. `isNull()` moved onto a new `PostgrestNullableExpression` refinement, reached by conditional conformance:

```swift
extension PostgrestStoredColumn: PostgrestNullableExpression
where Nullability == PostgrestNullable {}
```

Scope check was as you said — only this file, `PostgrestFilter+Comparison.swift` and the test files. Nothing macro-generated depends on the old shape yet.

**2. Corrected the `raw` docs.** You were right that the immediate junction is the wrong test. `queryItems(for:)` maps every child of an `.or` through `group()`, and `group(.and(…))` recurses, so ancestry is what decides it. Both of your examples confirmed:

```
(a && .raw("cost::text", "eq.10")) || b  →  or=(and(a,cost::text.eq.10),b)
!(a && .raw("cost::text", "eq.10"))      →  not.and=(a,cost::text.eq.10)
```

Both doc comments now state the ancestry rule, and both examples are pinned as tests in `PostgrestFilterTests.swift` — that nesting had no coverage.

**3. `PostgrestOperator` reused** for all eight (`is` needs backticks: `PostgrestOperator.`is`.rawValue`).

One thing beyond the three findings: `sdk-compliance.yaml` needed updating, since the members now live on the struct rather than the typealiases. Names came from a `swift package dump-symbol-graph` dump rather than guesswork — it confirmed the typealiases are emitted as symbols but `PostgrestColumn.init` is not, so those entries moved to `PostgrestStoredColumn`.

Source-compatible: `PostgrestColumn<Todo, Int>("id")` and `.isNull()` on a nullable column both still compile. `format.sh` clean, 195 PostgREST tests pass, full suite 1342, docs build clean.

Heads-up: this moves the tip of #1288, so #1289–#1299 need a rebase. I expect rebase-only, no edits — the typealiases keep `PostgrestNullableColumn<R, V>` valid as a type reference in `PostgrestUpdate`'s overloads and in the macro output — but I haven't confirmed that by actually rebasing yet.

spydon (approved):
Lgtm!

PR #1289, opened 2026-08-27
feat(postgrest): add the remaining filter operators

grdsdev:
Stack 2/8 for SDK-1624. Base: `guilhermesouza/sdk-1624-column-expressions`.

Pattern, list, array, range, jsonb and text-search operators, completing the operator surface.

Two details are wire-verified rather than inferred:

- `contains`/`containedBy`/`overlaps` each get three methods, because the operand literal is chosen by the column's Postgres type and not by the operator: `{a,b}` for an array, `[1,3)` for a range, `{"a":1}` for jsonb. One method taking one argument cannot spell all three.
- `in` renders its own `in.(…)` delimiters, which must reach the server literal. It therefore builds a `raw` node so `group()` leaves it alone — escaping it gives `or=(id.in."(2,3)",…)`, a 400. Members are still escaped.

`in` is declared as `` func `in` `` — the keyword needs backticks in the declaration, but **not at the call site**: `$0.id.in([1, 2])` parses, and so does `{ c in c.id.in([1, 2]) }` with both meanings of `in` on one line. The name matches the wire token, supabase-js, the legacy builder's `in(_:values:)`, and the `database.using_filters.in` capability id.

`regexMatch`/`regexIMatch` deliberately avoid the name `match`: supabase-js binds `match` to a multi-column equality shorthand, and the capability matrix has separate ids for the three meanings.

### DIFF
--- a/Sources/PostgREST/Filters/PostgrestFilter+Collection.swift
+++ b/Sources/PostgREST/Filters/PostgrestFilter+Collection.swift
@@ -1,0 +1,142 @@
+//
+//  PostgrestFilter+Collection.swift
+//  PostgREST
+//
+//  Created by Guilherme Souza on 26/08/26.
+//
+
+extension PostgrestFilter {
+  /// `{a,b}` — the Postgres array literal the array operators take.
+  static func array(_ values: [some PostgrestArrayElement]) -> String {
+    "{\(values.map(\.postgrestArrayElement).joined(separator: ","))}"
+  }
+}
+
+// MARK: - Array operands
+//
+// `cs`, `cd` and `ov` each get three methods, because the operand literal is chosen by the
+// column's Postgres type rather than by the operator: `{a,b}` for an array, `[1,3)` for a range,
+// `{"a":1}` for jsonb. Mixing them is a 400.
+//
+// Only the array shape is type-checked (`where Value == [E]`), so `contains(["a"])` on a
+// `String` column does not compile. A range and a jsonb column have no distinguishing Swift
+// type to constrain on, so the range and JSON methods sit on bare
+// `PostgrestFilterableExpression` and compile on any column; pairing one with the wrong column
+// is a server error (`42883 operator does not exist`), not a wrong answer.
+
+extension PostgrestFilterableExpression {
+  /// Matches rows where this array column contains every element of `values`.
+  ///
+  /// An empty `values` matches every row whose column is non-null — the opposite of `in` and
+  /// `likeAnyOf`, since every array contains the empty array.
+  public func contains<E: PostgrestArrayElement>(_ values: [E]) -> PostgrestFilter<Root>
+  where Value == [E] {
+    PostgrestFilter(
+      column: postgrestExpression, operator: "cs", value: PostgrestFilter<Root>.array(values))
+  }
+
+  /// Matches rows where every element of this array column is contained by `values`.
+  public func containedBy<E: PostgrestArrayElement>(_ values: [E]) -> PostgrestFilter<Root>
+  where Value == [E] {
+    PostgrestFilter(
+      column: postgrestExpression, operator: "cd", value: PostgrestFilter<Root>.array(values))
+  }
+
+  /// Matches rows where this array column shares at least one element with `values`.
+  public func overlaps<E: PostgrestArrayElement>(_ values: [E]) -> PostgrestFilter<Root>
+  where Value == [E] {
+    PostgrestFilter(
+      column: postgrestExpression, operator: "ov", value: PostgrestFilter<Root>.array(values))
+  }
+}
+
+// MARK: - Range operands
+//
+// A range literal is taken as a string, and must stay a `.comparison` so `group()` escapes its
+// `)`/`]`: bare, they close an enclosing `or=(…)` early and 400. The opposite of `in`.
+
+extension PostgrestFilterableExpression {
+  /// Matches rows where this range column contains `range`.
+  ///
+  /// - Parameter range: A Postgres range literal, for example `"[2,3)"`.
+  public func containsRange(_ range: String) -> PostgrestFilter<Root> {
+    PostgrestFilter(column: postgrestExpression, operator: "cs", value: range)
+  }
+
+  /// Matches rows where this range column is contained by `range`.
+  public func containedByRange(_ range: String) -> PostgrestFilter<Root> {
+    PostgrestFilter(column: postgrestExpression, operator: "cd", value: range)
+  }
+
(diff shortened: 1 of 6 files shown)


Tr00d (approved):

spydon (changes requested):
I stood up Postgres 15 + PostgREST v14.14 locally to check the wire claims rather than reason about them. Confirmed: `in` genuinely must be a raw node (`or=(id.in."(1,3)",…)` gives `PGRST100: unexpected "\"" expecting "("`), and group position tolerates more than the comments assume, `or=(name.like(all)."{a%,ab%}",…)` and `or=(name.fts(english).abc,…)` both return 200 with the right rows. `!$0.id.in(…)` also renders correctly in both positions.

**1. The empty-list doc claim on `likeAllOf`/`ilikeAllOf` is inverted** (`PostgrestFilter+Pattern.swift:57` and `:75`). It's accepted without error, but it matches *every* row, not none:

```
?name=like(all).{}   -> all 3 rows   [200]
?name=ilike(all).{}  -> all 3 rows   [200]
?name=like(any).{}   -> []           [200]
```

PostgREST renders these as SQL quantified comparisons and `ALL` over an empty array is vacuously true: `'abc' like all ('{}'::text[])` is `true`, and so is `null::text like all ('{}'::text[])`, so even NULL-column rows come back. Same vacuous-truth reasoning you already applied correctly to `contains`, so it looks like a slip. Worth fixing because the empty case is exactly what you hit when the pattern list comes from user input, and "matches nothing" vs "matches everything" is the difference between an empty screen and a full table dump. The `likeAnyOf`/`ilikeAnyOf` lines are correct as written, and `in.()` really does match nothing so that one's fine too.

**2. `braced(_:)` duplicates `array(_:)`** (`PostgrestFilter+Pattern.swift:22` vs `+Collection.swift:10`). `String` gets `postgrestArrayElement` from the default impl on `PostgrestFilterValue`, which is exactly `escapePostgRESTArrayLiteralElement(rawValue)`, so `array(patterns)` is byte-identical. Dropping `braced` removes the "which escaper does this one use" question at four call sites. The `list(_:)` vs `array(_:)` split is the load-bearing one and should stay.

**3. `database.using_filters.not_in` gets nothing** (`sdk-compliance.yaml:862`). Every other operator row picked up a `PostgrestFilterableExpression.*` symbol. The typed surface deliberately has no `notIn` and spells it `!$0.id.in([1, 2])`, which is a decision the matrix should record, especially since #1288 added a note to `.raw` for the same reason.

Minor:

- The new note on `using_filters.contains` says the string builder "takes one dictionary/array argument", but `contains(_:value:)` takes `any PostgrestFilterValue` and no `Dictionary` conformance ships, so jsonb containment there is a hand-written literal too.
- Stray blank line in the YAML: the new `ilikeAnyOf` entry landed after the blank line that separated `using_filters` from `using_modifiers`, so the separator is gone.
- `PostgrestFilter<Root>.array(…)`/`.list(…)` never touch `R`; as free internal functions the nine call sites lose the generic qualification.
- `textSearch` reusing `TextSearchType` is good, but it puts the new surface on a type under `Legacy/`. If that folder is headed for retirement it should move out (same applies to `PostgrestOperator`, so I'd hold off on the string-literal point I raised on #1288 until that's settled).

grdsdev:
Standing up Postgres + PostgREST to check the wire rather than the docs is the right call, and it caught a real inversion. Fixed in 2a03364.

**1. `likeAllOf`/`ilikeAllOf` empty-list doc — inverted, fixed.** `LIKE ALL('{}')` is vacuously true, so an empty list matches every row including `NULL`-column ones. The docs now say that, and name the `any` case as the reverse:

```swift
/// An empty `patterns` matches **every** row, including rows where the column is `NULL`:
/// PostgREST renders this as `LIKE ALL('{}')`, and a quantified `ALL` over an empty array is
/// vacuously true. `likeAnyOf`/`ilikeAnyOf` are the reverse — empty matches no rows.
```

Also added `emptyListsRenderAnEmptyLiteral()`, pinning `like(all).{}`, `like(any).{}` and `in.()`. The behaviour is the server's, but the render is ours, and the guard that matters is nobody later "helpfully" short-circuiting an empty list to a never-matching filter — which is exactly the bug the old comment described.

**2. `braced(_:)` duplicated `array(_:)` — dropped.** `String: PostgrestFilterValue` with `rawValue == self` (`PostgrestFilterValue.swift:57`), and the default `postgrestArrayElement` is `escapePostgRESTArrayLiteralElement(rawValue)` (`:50`), so the two were byte-identical. `list(_:)` stays; agreed that split is load-bearing.

**3. `not_in` records the decision now.**

```yaml
note: "The typed surface has no dedicated `notIn`: negation is the `!` operator on the filter
  tree, so `!$0.id.in([1, 2])` renders `id=not.in.(1,2)`."
```

Backed by an existing assertion at `PostgrestFilterPatternTests.swift:69`, so the note can't drift from the render.

**Minors:**

- Blank line restored above `using_modifiers.order`.
- `list`/`array` are now free internal `postgrestList`/`postgrestArray`; nine call sites lose the `PostgrestFilter<Root>.` qualification. Kept the `postgrest` prefix rather than bare `array`/`list` — too generic for module scope.
- **`using_filters.contains` note — pushing back, the note is accurate.** `JSONObject` is `[String: JSONValue]` (`JSONValue.swift:3`) and *does* conform to `PostgrestFilterValue` with a JSON-literal `rawValue` (`PostgrestFilterValue.swift:147`), so the dictionary form works on the string builder without a hand-written literal. There's a shipped test for it: `.contains("address", value: ["postcode": 90210])` at `PostgrestFilterBuilderTests.swift:129`. Happy to reword if "dictionary/array argument" reads as claiming more than that.
- **`TextSearchType`/`PostgrestOperator` under `Legacy/`** — agreed this needs settling before either surface leans on that folder. One thing to flag: I'd already landed the `PostgrestOperator` string-literal change on #1288 before this review, so it's in `0bc1a30`. Leaving it rather than churning it back — the references survive a move out of `Legacy/` and only break on outright retirement, which is the conversation you're pointing at. Say the word if you'd rather it come out now.

Gates on `2a03364`: format clean, 1360 tests pass, docs build clean, YAML parses.

One correction on process, not code: my first push of this fix went out on a stale base and briefly clobbered the rebase that was already on the remote. Restored by resetting to `aa6f46ec` and cherry-picking, so the branch is back on `0bc1a30` — the current tip of #1288 — with only the four review-fix files on top. Nothing of yours or #1288's was lost, but the force-push in between was mine and shouldn't have happened.

spydon (approved):
Lgtm!

PR #1290, opened 2026-08-27
feat(postgrest): generate a Columns namespace from @Table

grdsdev:
Stack 3/8 for SDK-1624. Base: `guilhermesouza/sdk-1624-filter-operators`.

`@Table` now emits a nested `Columns` struct holding one column value per stored property, and `PostgrestRelation` gains `Columns`/`columns` so callers can reach it generically.

An optional property gets a `PostgrestNullableColumn` carrying its **wrapped** type (`var dueDate: Date?` → `PostgrestNullableColumn<Todo, Date>`). That is what lets an update accept `nil` for a nullable column and reject it for a `NOT NULL` one, while keeping every operator at one declaration — no operator takes an optional operand.

Two existing readers move onto it:

- `PostgrestUpdate`'s subscript keys off `KeyPath<R.Columns, …>` instead of the row type, so an assignment resolves the column name from the same place a filter does.
- `@SelectionOf` builds `selectString` from the namespace.

`upsert(onConflict:)` follows. It used `columnName(for:)`, which this stack removes, so it moves to key paths into `Columns` — with parameter packs, so `\.email` still infers and existing call sites are untouched.

`columnName(for:)` stays for now: the key-path filter surface still reads it, and it goes in 5/8.

### DIFF
--- a/Sources/PostgREST/Query/PostgrestTypedMutation.swift
+++ b/Sources/PostgREST/Query/PostgrestTypedMutation.swift
@@ -84,6 +84,11 @@ extension PostgrestTypedSource where R: PostgrestWritableRelation {
   /// namespace, so a column that does not exist is a compile error rather than a PostgREST 400.
   /// Taking a first column plus the rest also makes an empty target unrepresentable.
   ///
+  /// Each key path must land on a ``PostgrestStoredColumn`` of *this* relation, which is narrower
+  /// than a column expression in general: `on_conflict` names columns of a unique index, so a
+  /// derived expression — a cast, an aggregate — is not a target PostgREST can take, and neither
+  /// is a column belonging to some other relation. Both are rejected at the call site.
+  ///
   /// To merge on the primary key, use ``upsert(_:)``, which derives the target instead.
   ///
   /// - Parameters:
@@ -93,12 +98,16 @@ extension PostgrestTypedSource where R: PostgrestWritableRelation {
   /// - Returns: A ``PostgrestTypedMutation`` to execute, or to request rows back from.
   /// - Throws: An encoding error if `values` cannot be serialized.
   public func upsert<
-    First: PostgrestColumnExpression,
-    each Rest: PostgrestColumnExpression
+    FirstValue,
+    FirstNullability: PostgrestNullability,
+    each RestValue,
+    each RestNullability: PostgrestNullability
   >(
     _ values: R.Draft,
-    onConflict column: KeyPath<R.Columns, First>,
-    _ additional: repeat KeyPath<R.Columns, each Rest>
+    onConflict column: KeyPath<R.Columns, PostgrestStoredColumn<R, FirstValue, FirstNullability>>,
+    _ additional: repeat KeyPath<
+      R.Columns, PostgrestStoredColumn<R, each RestValue, each RestNullability>
+    >
   ) throws -> PostgrestTypedMutation<R> {
     var names = [R.columns[keyPath: column].postgrestExpression]
     repeat names.append(R.columns[keyPath: each additional].postgrestExpression)
--- a/Sources/PostgREST/Query/PostgrestTypedMutation.swift
+++ b/Sources/PostgREST/Query/PostgrestTypedMutation.swift
@@ -80,10 +80,9 @@ extension PostgrestTypedSource where R: PostgrestWritableRelation {
   ///   .execute()
   /// ```
   ///
-  /// The target is spelled as key paths, so a column that does not exist is a compile error rather
-  /// than a PostgREST 400 naming a column you cannot grep for. Splitting it into a first column
-  /// plus the rest also makes an empty target unrepresentable — `on_conflict=` with nothing after
-  /// it is a different request, not a missing one.
+  /// The target is spelled as key paths into the relation's ``PostgrestRelation/Columns``
+  /// namespace, so a column that does not exist is a compile error rather than a PostgREST 400.
+  /// Taking a first column plus the rest also makes an empty target unrepresentable.
   ///
   /// To merge on the primary key, use ``upsert(_:)``, which derives the target instead.
   ///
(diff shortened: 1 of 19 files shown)


Tr00d (approved):

spydon (changes requested):
The shape here is right. `Columns` as a struct of instance `let`s (rather than static members) is what makes `KeyPath<R.Columns, …>` work, the explicit `public init() {}` is genuinely required and correctly explained, and extending the `@Table` `names:` list is the kind of thing that's easy to forget and produces a baffling error when you do. Three things:

**1. An implicitly-unwrapped-optional property now fails inside macro-generated code.** `postgrestStoredProperties()` stores the annotation verbatim (`StoredProperty.swift:103`) and `postgrestIsOptionalType` only recognises a `?` suffix or an `Optional<` prefix, so `var title: String!` gets `isOptional == false` and `columnsNamespace` emits it straight into a generic argument (`TableMacro.swift:178`):

```swift
public let title = PostgrestColumn<Todo, String!>("title")
```

which doesn't compile: `error: using '!' is not allowed here / note: use '?' instead`. The error lands on expanded code, so the author sees it pointed at a line they didn't write. It's a regression rather than a pre-existing gap: `optionalType` is only consulted for `isOptional || hasDefault` (`TableMacro.swift:120`), so a plain `var title: String!` previously flowed through as a property type, where `!` *is* legal, and worked. (`@Default var title: String!` was already broken, since that builds `String!?`.) Treating a `!` suffix as optional is probably the right call since that's what the spelling means for a nullable column, and it makes a diagnostic unnecessary.

Related root cause, worth fixing in the same place: `postgrestIsOptionalType` and `unwrappedType` decide optionality with *different* predicates. The first accepts `hasPrefix("Optional<")` alone; the second requires the prefix *and* `hasSuffix(">")`. Any spelling satisfying one and not the other (`Optional<Int>!` being the reachable one) yields `PostgrestNullableColumn<Todo, Optional<Int>>`, whose `Value` is itself optional, which is the one thing the wrapped-type design exists to prevent. And it fails quietly: `Optional` doesn't conform to `PostgrestFilterValue`, so every operator just disappears from the column and you get "no member `eq`". Deriving both from one function would close the class rather than the instance.

**2. The breaking surface is narrow but real, and the commit isn't marked breaking.** `PostgrestRelation` gains `Columns` and `columns` with no defaults. `@Table` users see nothing change, including `$0.task = …`, since `dynamicMemberLookup` now resolves against `Columns`. But `PostgrestRelation.swift` advertises hand-written conformances as the supported escape hatch, and those break three ways at once: two new requirements, the `KeyPath<R, V>` update subscript gone, and `@SelectionOf` now resolving through `Columns` instead of `columnName(for:)` (so a hand-written relation needs a hand-written `Columns` with a member per selected property, not just a working `columnName(for:)`). There's also one niche break for generated relations: `upsert(onConflict: \Todo.email)` with an explicit root no longer typechecks, since the root is now `Todo.Columns`. The bare `\.email` spelling does still infer, as you say.

Precedent is `refactor!: remove all deprecated pre-v3 APIs (#1206)`, so `feat(postgrest)!` looks right. If the intent is that everything in the 3.0.0-beta window is implicitly breaking and the stack deliberately skips `!`, that's a fine answer, I just want to know it's deliberate.

**3. `upsert`'s pack bound admits expressions `on_conflict` can't take.** `First: PostgrestColumnExpression` is the *select-position* protocol, so once #1293/#1294 land it also covers casts and aggregates, neither of which is a valid `on_conflict` target. Nothing exposes those through `Columns` today so it's latent, but the bound is wider than the operation. Neither `First` nor `each Rest` is constrained to `Root == R` either, which is the guarantee the rest of the design leans on; if a stored-column marker protocol isn't worth adding, `Root == R` at least restores the relation check.

Minor: the compliance diff adding only `PostgrestRelation.Columns`/`.columns` is correct, the generated members live in the user's module. And `columnsNamespace` correctly uses `property.columnName` for the wire name and `property.name` for the Swift member, so `@Column("…")` carries through — worth watching when `columnName(for:)` is deleted in 5/8, since that's the point where the two mappings stop cross-checking each other.

grdsdev:
All three addressed in aa7e6d9.

**1. The IUO regression — reproduced, and it is a regression.** `var title: String!` on the parent commit `ac69c0ce`: builds clean. On `fa951812`: `macro expansion @Table:10:49: error: using '!' is not allowed here`, pointed at a line the author never wrote, exactly as described.

Both halves come from one place now, since the two predicates disagreeing was the actual class of bug:

```swift
func postgrestUnwrapOptionalType(_ type: String) -> String? {
  let inner: String
  if type.hasSuffix("?") || type.hasSuffix("!") {
    inner = String(type.dropLast())
  } else if type.hasPrefix("Optional<"), type.hasSuffix(">") {
    inner = String(type.dropFirst("Optional<".count).dropLast())
  } else {
    return nil
  }
  return postgrestUnwrapOptionalType(inner) ?? inner
}

func postgrestIsOptionalType(_ type: String) -> Bool {
  postgrestUnwrapOptionalType(type) != nil
}
```

Two decisions worth flagging, since neither is forced:

- **`!` is treated as nullable**, so no diagnostic is needed — agreed that's what the spelling means for a column. It also fixes `@Default var title: String!`, which built `String!?` before.
- **Every layer is stripped, not one.** Stripping one still yields `PostgrestNullableColumn<Todo, Int?>` for `Optional<Int>!`, which is the same silent "no member `eq`" you identified, just reached differently. A column is nullable or it is not, so `Optional<Int>!` and `Int??` both mean a nullable `Int`, and full stripping makes "`Value` is never optional" hold by construction rather than by case analysis.

Coverage: `optionalityAgreesWithUnwrapping()` walks both functions over the same spellings so they cannot drift apart again, `unwrapsEveryOptionalSpelling()` pins `String!`, `Optional<Int>!` and `Int??`, and `IntegrationDraftNote` carries a real `var title: String!` — that one is compile-time coverage, which is what the failure mode actually was.

**2. No `!` is deliberate.** Every layer of this stack is implicitly breaking and none of them carry the marker. The escape hatch you're pointing at has never shipped: `Sources/PostgREST/Relations/PostgrestRelation.swift` does not exist in `v2.55.1`, `v2.55.0` or `v2.54.1`, so there is no released hand-written conformance to break, and the same goes for the `KeyPath<R, V>` update subscript and the explicit-root `upsert(onConflict: \Todo.email)` spelling. `#1206` was different — that removed APIs people were actually using.

**3. Bound narrowed, and the relation check is back in both positions.** No marker protocol needed — since #1288 collapsed the two column structs into one, the concrete type says it directly:

```swift
public func upsert<
  FirstValue,
  FirstNullability: PostgrestNullability,
  each RestValue,
  each RestNullability: PostgrestNullability
>(
  _ values: R.Draft,
  onConflict column: KeyPath<R.Columns, PostgrestStoredColumn<R, FirstValue, FirstNullability>>,
  _ additional: repeat KeyPath<
    R.Columns, PostgrestStoredColumn<R, each RestValue, each RestNullability>
  >
) throws -> PostgrestTypedMutation<R>
```

This gets `Root == R` in the pack position too, which `repeat (each Rest).Root == R` cannot express — that's "same-element requirements are not yet supported", which is what made me drop the constraint in the first place. Zipping two packs of matching shape sidesteps it.

Verified by compile probe against a hand-written `Columns` carrying a `FakeExpression: PostgrestColumnExpression` and a `PostgrestColumn<ProbeOther, Int>`. All four rejected — `\.aggregate`, `\.foreign`, and each of those in the pack position — while `onConflict: \.email` and `onConflict: \.email, \.id` still infer. The foreign-column error names the real problem:

```
note: arguments to generic parameter 'Root' ('ProbeOther' and 'ProbeUser') are expected to be equal
```

The probes are throwaway, not committed; the positive spellings are already covered by the existing upsert tests, and there's no negative-compile harness in this repo to hang the rest on.

On the closing minor: agreed, and `columnName(for:)` going away in 5/8 is the point where `property.columnName` vs `property.name` stops being cross-checked. Watching for it there.

Gates on `aa7e6d9`: 1367 tests pass, format clean, docs build clean.

spydon (approved):

PR #1291, opened 2026-08-27
feat(postgrest): add the where and order closures

grdsdev:
Stack 4/8 for SDK-1624. Base: `guilhermesouza/sdk-1624-columns-namespace`.

```swift
try await client.from(Todo.self)
  .select()
  .where { ($0.isDone.eq(false) && $0.priority.gt(3)) || $0.id.eq(7) }
  .order { $0.dueDate }
  .execute()
```

`where(_:)` takes a closure over the column namespace and returns a `PostgrestFilter`, so a whole boolean expression is one call instead of a chain of per-operator methods. Because the filter is a value, one that depends on runtime state is assembled in the closure body like any other value.

`order(_:)` is constrained to `PostgrestOrderableExpression` — a third position set that overlaps the filterable set without matching it. `order=cost::text.desc` (a cast) and `order=children(amount.sum()).desc` (an aggregate) are both 400s on PostgREST 16.1, so neither compiles.

The direction is optional: `order { $0.dueDate }` sends no direction and lets PostgREST apply its own default. `.asc()`/`.desc()` are for when it matters, and chain into `nullsFirst()`/`nullsLast()`.

Existing key-path call sites move over here; the old surface is deleted in 5/8.

### DIFF
--- a/Sources/PostgREST/Columns/PostgrestOrdering.swift
+++ b/Sources/PostgREST/Columns/PostgrestOrdering.swift
@@ -32,17 +32,14 @@ public struct PostgrestOrdering<Root: PostgrestRelation>: Sendable {
       + (nullPlacement.map { ".\($0.rawValue)" } ?? "")
   }
 
-  /// Sorts `NULL`s before non-null values.
-  public func nullsFirst() -> Self {
+  /// Places `NULL`s relative to non-null values.
+  ///
+  /// ```swift
+  /// .order { $0.priority.desc().nulls(.first) }
+  /// ```
+  public func nulls(_ placement: PostgrestNullPlacement) -> Self {
     var copy = self
-    copy.nullPlacement = .first
-    return copy
-  }
-
-  /// Sorts `NULL`s after non-null values.
-  public func nullsLast() -> Self {
-    var copy = self
-    copy.nullPlacement = .last
+    copy.nullPlacement = placement
     return copy
   }
 }
@@ -65,8 +62,7 @@ public struct PostgrestNullPlacement: RawRepresentable, Hashable, Sendable {
 extension PostgrestOrderableExpression {
   /// Sorts by this expression, smallest first.
   ///
-  /// Null placement is left to the database unless you chain
-  /// ``PostgrestOrdering/nullsFirst()`` or ``PostgrestOrdering/nullsLast()``.
+  /// Null placement is left to the database unless you chain ``PostgrestOrdering/nulls(_:)``.
   public func asc() -> PostgrestOrdering<Root> {
     PostgrestOrdering(column: postgrestExpression, ascending: true)
   }
@@ -74,5 +70,19 @@ extension PostgrestOrderableExpression {
   /// Sorts by this expression, largest first.
   public func desc() -> PostgrestOrdering<Root> {
     PostgrestOrdering(column: postgrestExpression, ascending: false)
+  }
+
+  /// Sorts by this expression with an explicit `NULL` placement, leaving the direction to
+  /// PostgREST.
+  ///
+  /// ```swift
+  /// .order { $0.dueDate.nulls(.first) }   // order=due_at.nullsfirst
+  /// ```
+  ///
+  /// The two parts are independent on the wire, so a placement does not require picking a
+  /// direction: `order=name.nullsfirst` is a 200. Without this, wanting PostgREST's default
+  /// direction *and* an explicit placement would force a direction into the query anyway.
+  public func nulls(_ placement: PostgrestNullPlacement) -> PostgrestOrdering<Root> {
+    PostgrestOrdering(column: postgrestExpression, ascending: nil, nullPlacement: placement)
   }
 }
--- a/Sources/PostgREST/Columns/PostgrestOrdering.swift
+++ b/Sources/PostgREST/Columns/PostgrestOrdering.swift
@@ -1,0 +1,78 @@
+//
+//  PostgrestOrdering.swift
+//  PostgREST
+//
+//  Created by Guilherme Souza on 26/08/26.
+//
+
+/// One sort key: a column expression, optionally with a direction and a null placement.
+///
+/// ```swift
+/// .order { $0.priority.desc() }
+/// .order { $0.id }              // breaks ties in the first, ascending by default
+/// ```
+///
+/// Omitting the direction sends `order=id` bare, so PostgREST's own default applies rather than
+/// a value frozen into the client.
+public struct PostgrestOrdering<Root: PostgrestRelation>: Sendable {
+  let column: String
+
+  /// `nil` sends no direction, so PostgREST's default applies (ascending, `NULL`s last on 16.1).
+  let ascending: Bool?
+
+  /// `nil` sends no placement, so the database default applies.
+  ///
+  /// Deliberately unlike `PostgrestTransformBuilder.order(_:ascending:nullsFirst:)`, which always
+  /// appends one — see [SDK-1633](https://linear.app/supabase/issue/SDK-1633).
+  var nullPlacement: PostgrestNullPlacement?
+
+  var rendered: String {
+    column
+      + (ascending.map { ".\($0 ? "asc" : "desc")" } ?? "")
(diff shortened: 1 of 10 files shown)


Tr00d (approved):

spydon (changes requested):
Ran the wire claims against Postgres 15 + PostgREST v14.14 locally. Confirmed, and one of these is a genuinely good call:

- `appendingOrder`'s merge into a single `order` parameter is **necessary**, not just tidy. PostgREST honours only the *first* `order` parameter and silently ignores the rest: `order=name.asc&order=id.desc` sorts by name only, `order=id.desc&order=name.asc` sorts by id only, and only the merged `order=name.asc,id.desc` applies both. The current comment explains why `builder.order(_:ascending:nullsFirst:)` isn't reused, which is the weaker half of the reason; "a second `order` parameter is silently ignored" is the part a future reader needs.
- `order=cost::text.desc` really is a 400 (`PGRST100 … failed to parse order`), so the `PostgrestOrderableExpression` constraint earns its keep.
- "Two `where` calls accumulate — they are ANDed" holds in the hard case: `or=(id.eq.1,id.eq.2)&or=(id.eq.2,id.eq.3)` returns only id 2.

I also checked the thing I expected to be broken and it isn't: #1295's `PostgrestToOneColumn.postgrestExpression` renders `owner(label)`, and `order=owner(label)` returns 200 and sorts the outer rows correctly (`order=owner.label` is a 400). Dropping `PostgrestFilterableExpression` from embedded columns while keeping `PostgrestOrderableExpression` is exactly right.

Three things:

**1. Null placement is unreachable without freezing a direction, and the server doesn't require that.** `nullsFirst()`/`nullsLast()` live on `PostgrestOrdering` (`PostgrestOrdering.swift:36`, `:43`), which only `asc()`/`desc()` produce; the bare-column `order` overload builds its `PostgrestOrdering` inline and renders it immediately (`+Where.swift:80`), so there's no value to chain onto. But PostgREST accepts a placement with no direction:

```
?order=name.nullsfirst   -> NULL row first, 200
?order=name.nullslast    -> NULL row last, 200
```

and `PostgrestOrdering.rendered` already produces exactly that when `ascending == nil` and `nullPlacement != nil`, since both parts are independently optional. So the capability exists in the type and on the server, only the API withholds it. The consequence is the specific thing this PR argues against elsewhere: the `ascending: Bool?` doc says `nil` is there so PostgREST's default applies "rather than a value frozen into the client", yet a caller who wants the default direction *and* explicit NULL placement has to pick a direction anyway. The doc at `+Where.swift:74-75` ("a placement hangs off the direction, not off the bare column") reads like a server fact and it's a client-side choice. Two methods on `PostgrestOrderableExpression` returning `PostgrestOrdering(column:ascending: nil, nullPlacement:)` closes it.

**2. `PostgrestNullPlacement` is public API nothing can reach.** `PostgrestOrdering`'s `column`, `ascending`, `nullPlacement`, `rendered` and memberwise init are all internal, and no public function anywhere accepts or returns a `PostgrestNullPlacement` — you can construct one and then have nowhere to put it. The compliance file still registers five symbols for it (`PostgrestNullPlacement`, `.init`, `.rawValue`, `.first`, `.last`), five things to support forever for zero reach. Either make it internal until something takes it, or let it *be* the parameter of the entry point in #1 (a single `nulls(_:)` instead of two methods). Unlike `PostgrestOperator` and `TextSearchType`, the value set here is closed by PostgREST's grammar, so RawRepresentable-for-extensibility buys nothing an enum wouldn't.

**3. `PostgrestKeyPathFilterable` is now the wrong name.** This PR adds `where(_:)` to it and 5/8 deletes every key-path method from it, leaving a protocol called "KeyPathFilterable" whose only member is a closure-based `where` that deliberately doesn't use key paths into the row type. It's public and registered in the compliance file (`PostgrestKeyPathFilterable.where` is added here), so renaming after the fact is another breaking change stacked on 1292's. Renaming here or in 1292 costs nothing.

Minor: three `order` overloads coexist until 1292 lands. I checked and there's no ambiguity (`PostgrestOrdering` doesn't conform to `PostgrestOrderableExpression`, and a closure never matches the key-path overload), but `.order(\.dueDate)` emits `due_at.asc.nullslast` while `.order { $0.dueDate.asc() }` emits `due_at.asc`, so the two spellings that look equivalent aren't, for as long as both exist.

grdsdev:
All three addressed in b006b7e.

**1 + 2 together — the placement is now the entry point's parameter.** These were one problem: the capability was unreachable, and the type that expresses it had nowhere to be used. Taking your second option closes both with less API than fixing them separately:

```swift
extension PostgrestOrderableExpression {
  /// Sorts by this expression with an explicit `NULL` placement, leaving the direction to
  /// PostgREST.
  public func nulls(_ placement: PostgrestNullPlacement) -> PostgrestOrdering<Root> {
    PostgrestOrdering(column: postgrestExpression, ascending: nil, nullPlacement: placement)
  }
}
```

`nullsFirst()`/`nullsLast()` on `PostgrestOrdering` collapse into one `nulls(_:)` as well, so there is a single spelling rather than `desc().nullsFirst()` next to `dueDate.nulls(.first)`. Net public methods: two, down from two, with `PostgrestNullPlacement` now load-bearing instead of unreachable. `aPlacementDoesNotRequireADirection()` pins `order=due_at.nullslast` and asserts no direction leaked in.

The misleading doc is gone too — "a placement hangs off the direction, not off the bare column" was stating a client-side choice as a server fact, which is exactly the thing the `ascending: Bool?` doc argues against.

**One pushback, on the enum sub-point.** Keeping `PostgrestNullPlacement` as a `RawRepresentable` struct rather than an enum. `AGENTS.md` draws the line at the wire, not at whether the value set looks closed:

> Use a Swift `enum` only when the value is genuinely closed: every case is fixed by the language or protocol itself (HTTP methods, a `CodingKeys` set) or the value is entirely client-side and never crosses the wire.
>
> Any enum-like value sent to or received from the backend must be a `RawRepresentable` struct instead […]

`nullsfirst`/`nullslast` are sent to the backend and are not fixed by Swift, so the rule lands on the struct side. I agree the extensibility buys little here in practice; the reason to keep it is consistency with `PostgrestOperator` and `TextSearchType` under a written project convention, not a prediction that PostgREST grows a third placement. Worth changing the convention if you think it's drawn in the wrong place, but not worth this type being the exception to it.

**3. Renamed to `PostgrestFilterableRequest`.** Checked what the protocol actually requires — `Relation`, `Phase`, `builder`, `init(builder:)` — and not one of them mentions a key path, so the old name was already only describing the extension methods that 5/8 deletes. `…Request` rather than bare `…Filterable` to keep it distinct from `PostgrestFilterableExpression`, which is a column, not a request. The stale doc on `Relation` ("the relation whose key paths this wrapper accepts") is fixed too.

Renamed here rather than in 1292 so the review resolves on the PR it was raised against. Two knock-ons: 1292 edits the same file and will want a trivial conflict resolution, and the compliance entries move with it (`PostgrestFilterableRequest.where`, plus the nine operator rows). No user-facing break — `PostgrestTypedQuery+Filters.swift` is absent from `v2.55.1` and `v2.55.0`, so nothing released names the old protocol.

**On `appendingOrder`:** the comment now leads with the part you flagged as load-bearing — only the first `order` parameter is honoured, the rest are silently ignored, so `order=name.asc&order=id.desc` sorts by name alone. That's the fact a future reader needs; "not `builder.order(_:ascending:nullsFirst:)`" is kept as the secondary note.

**On the minor:** `.order(\.dueDate)` emitting `due_at.asc.nullslast` against `.order { $0.dueDate }` emitting `due_at` is real, and it's exactly what makes the key-path overload worth removing rather than keeping as a shorthand. Left as-is since 1292 deletes it; if that PR slips, this one is the divergence to point at.

Gates on `b006b7e`: 1380 tests pass, format clean, docs build clean, YAML parses.

Note on the base: #1289 and #1290 merged while I was working, and this branch was rebased onto `845ff606` externally. Trees were identical, so the fix moved across unchanged, and I re-ran the full suite on the new base before pushing.

spydon (approved):

PR #1295, opened 2026-08-27
feat(postgrest): add embedded relations to the column namespace

grdsdev:
Stack 8/8 for SDK-1624. Base: `main`.

`PostgrestToOneRelation`/`PostgrestToManyRelation` sit on a relation's `Columns` namespace and project a column through the embed: `orders(total)` in a `select` list. They come from the postgres-meta-backed generator or a hand-written namespace — `@Table` cannot synthesize them, since a macro cannot see foreign-key metadata.

Cardinality reaches the type system because the server distinguishes it: `order=parent(id).asc` is fine, while the to-many direction is PGRST118 "do not form a many-to-one or one-to-one relationship". So a to-one projection is orderable and a to-many projection is not.

An embedded projection renders `parent(title)`, so a cast, JSON path or aggregate on one has to land **inside** those parentheses — PostgREST 16.1 answers 200 and silently drops anything outside them. Two pieces carry that rule, instead of each embedded column type shadowing the eight accessors it inherits:

- `PostgrestColumnExpression` gains one hook, `_deriving(_:)`, which places a derivation where the server expects it. Non-embedded expressions get a default that appends; each embed kind overrides it once.
- A phantom `Position` associated type carries where an expression may appear, so a derivation inherits its receiver's positions instead of restating them. `PostgrestDerivedExpression` conforms to the filterable and orderable protocols conditionally on it, which makes a derivation's position the intersection of the accessor's and the receiver's.

One `PostgrestDerivedExpression` therefore replaces `PostgrestCastColumn`, `PostgrestJSONPath` and `PostgrestAggregate`, and no fifth type is needed to keep a to-one chain inside the parentheses. The eight accessors are declared once each: 32 public accessor declarations across `Columns/` become 11, and 24 shadows become 0.

`PostgrestCastColumn` and `PostgrestAggregate` stay as aliases, since a cast and an aggregate are select-only whatever they are applied to. There is no `PostgrestJSONPath` alias: a JSON path inherits its receiver's position, so there is no `Position` to fix it to.

This also tightens one guarantee over the flattened design: a JSON path chained onto a cast now inherits the cast's select-only position, so filtering `cost::text->>k` — a 400 — no longer compiles.

The embed name is exposed as `postgrestEmbedName`, not `name`. Every public member of the relation type shadows a projected column spelled the same way, because `@dynamicMemberLookup` only fires when no real member matches — with a plain `name`, `Order.columns.todo.name` type-checked as the embed's own name string and the `name` column was unreachable. That is a silently wrong `select`, not an error.

Filtering *inside* an embed is still out of scope (SDK-1575); `embeddedFilterName` is what it will read.

---

Supersedes #1299, which proposed this design as a follow-up refactor. Folding it in here means the 24 shadows are never written in the first place.

### DIFF
--- a/Sources/PostgREST/Columns/PostgrestAggregate.swift
+++ b/Sources/PostgREST/Columns/PostgrestAggregate.swift
@@ -5,11 +5,17 @@
 //  Created by Guilherme Souza on 26/08/26.
 //
 
-/// An aggregate function applied to an expression — **select position only**.
+extension PostgrestDerivedExpression where Value == Int, Position == PostgrestSelectOnly {
+  /// `count()` — counts rows rather than values of a column.
+  public static var countAll: Self { Self(embed: nil, inner: "count()") }
+}
+
+/// The five aggregate functions, each declared once and correct for a stored column, a JSON path
+/// and either embed direction alike.
 ///
-/// PostgREST has no `HAVING`, and rejects an aggregate in `order`, so filtering or ordering by
-/// one is a compile error. Grouping needs nothing declared: selecting a plain column alongside an
-/// aggregate groups by it.
+/// The result is **select position only**: PostgREST has no `HAVING`, and rejects an aggregate in
+/// `order`, so filtering or ordering by one is a compile error. Grouping needs nothing declared —
+/// selecting a plain column alongside an aggregate groups by it.
 ///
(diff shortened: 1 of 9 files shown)


spydon (approved):
Lgtm!

PR #1299, opened 2026-08-28
refactor(postgrest): collapse the embedded-column accessor shadows

grdsdev:
Stack 9/9 for SDK-1624. Base: `guilhermesouza/sdk-1624-embedded-relations` (#1295).

> **Why not on top of #1290, as asked.** Three of the four files this rewrites do not exist there: `PostgrestDerivedColumn.swift` arrives in #1293, `PostgrestAggregate.swift` in #1294, `PostgrestRelatedColumns.swift` in #1295. At #1290 there is nothing to collapse. See "Alternative" at the bottom for the version that *does* start at #1290.

## The problem

Each embedded column type shadows the eight accessors it inherits from `PostgrestColumnExpression`, because an embedded projection renders `parent(title)` and a cast or aggregate has to land **inside** those parentheses — PostgREST answers 200 and silently drops anything outside. That is **24 declarations across three types**, and nothing makes a ninth accessor's missing shadows an error: it would be inherited, render outside, and return wrong rows quietly.

## The change

The shadows exist because a derived expression stores an already-flattened string, so nothing can be inserted after the fact.

```swift
// one hook: place a derivation where the server expects it
func _deriving<V, P: PostgrestPosition>(_ d: String) -> PostgrestDerivedExpression<Root, V, P>
// default appends; each embed kind overrides it once

// a phantom Position, so a derivation inherits its receiver's positions
associatedtype Position: PostgrestPosition
extension PostgrestDerivedExpression: PostgrestOrderableExpression
  where Position: PostgrestOrderablePosition {}
```

One `PostgrestDerivedExpression` replaces `PostgrestCastColumn`, `PostgrestJSONPath`, `PostgrestAggregate` and `PostgrestToOneDerivedColumn`. The last existed *only* to keep a chain inside the parentheses; that now falls out of the hook.

| | before | after |
| --- | --- | --- |
| shadows in `PostgrestRelatedColumns.swift` | 24 | 2 |
| public accessor declarations in `Columns/` | 32 | 11 |
| public types for derived expressions | 4 | 1 |

## Verification

- **1397 tests pass.** One test changed, and it *tightens* a guarantee: a JSON path chained onto a cast now inherits the cast's select-only position, so filtering `cost::text->>k` — a 400 the current design lets you write — no longer compiles.
- **Position matrix re-checked by compile probe.** Rejected: cast-of-a-to-one isn't orderable; to-many JSON path isn't orderable; to-many column isn't orderable; aggregate isn't filterable; to-one column isn't filterable; JSON-path-on-a-cast isn't filterable (new). Compiles: to-one JSON path is orderable; plain JSON path is filterable.
- **Drift scenario demonstrated.** Added a ninth accessor as one declaration with no shadows: it rendered `parent(name.lower())` and `kids(name.lower())`, and composed to `parent(name.lower()::text)`.

## Costs, stated plainly

- Merges four public types into one generic. 37 lines of `sdk-compliance.yaml` reference symbols this changes (not yet updated here).
- `_deriving(_:)` is public-for-protocol-reasons — underscored, but visible.
- Diagnostics regress in one way: a mispositioned expression fails as "does not conform to `PostgrestOrderableExpression`" rather than "has no member `asc`".
- As stacked, the reviewer reads "add 24 shadows" in #1293–#1295 and "delete them" here.

## Alternative, if that last point bothers you

Fold the `Position`/`_deriving` machinery into #1288 and re-author #1293/#1294/#1295 on top of it, so the shadows are never written. That is the version that begins near #1290. It is strictly better history and strictly more work — three pushed PRs rewritten instead of one added. Happy to do it that way instead; this PR is the cheap way to evaluate the design first.

### DIFF
--- a/Sources/PostgREST/Columns/PostgrestAggregate.swift
+++ b/Sources/PostgREST/Columns/PostgrestAggregate.swift
@@ -5,11 +5,17 @@
 //  Created by Guilherme Souza on 26/08/26.
 //
 
-/// An aggregate function applied to an expression — **select position only**.
+extension PostgrestDerivedExpression where Value == Int, Position == PostgrestSelectOnly {
+  /// `count()` — counts rows rather than values of a column.
+  public static var countAll: Self { Self(embed: nil, inner: "count()") }
+}
+
+/// The five aggregate functions, each declared once and correct for a stored column, a JSON path
+/// and either embed direction alike.
 ///
-/// PostgREST has no `HAVING`, and rejects an aggregate in `order`, so filtering or ordering by
-/// one is a compile error. Grouping needs nothing declared: selecting a plain column alongside an
-/// aggregate groups by it.
+/// The result is **select position only**: PostgREST has no `HAVING`, and rejects an aggregate in
+/// `order`, so filtering or ordering by one is a compile error. Grouping needs nothing declared —
+/// selecting a plain column alongside an aggregate groups by it.
 ///
 /// > Important: Requires PostgREST's `db-aggregates-enabled` setting. It is on for hosted
 /// > Supabase and off by default when self-hosting.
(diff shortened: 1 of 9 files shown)


grdsdev:
Folded into #1295 rather than landing as a follow-up, so the 24 accessor shadows are never written in the first place. Nothing is lost — #1295 now carries the `Position`/`_deriving(_:)` machinery and the collapsed `PostgrestDerivedExpression` as one commit.

Two things changed in the fold:

- The prototype was written against the pre-review version of `PostgrestColumnExpression.swift`, where `PostgrestColumn` and `PostgrestNullableColumn` were separate structs. #1288's review collapsed those into `PostgrestStoredColumn<Root, Value, Nullability>`, so the merged version keeps that plus `PostgrestNullableExpression`, with `PostgrestStoredColumn` gaining `Position = PostgrestEveryPosition`.
- `sdk-compliance.yaml` is updated, which the prototype never did: `PostgrestJSONPath` and `PostgrestToOneDerivedColumn` deregistered, `PostgrestDerivedExpression` and the six `Position` types registered, and the 24 per-type accessor entries replaced by the eight declarations on `PostgrestColumnExpression`. Names verified against a `swift package dump-symbol-graph` — 41 checked, 0 missing. That dump is also what showed `countAll` now belongs to `PostgrestDerivedExpression` rather than to the `PostgrestAggregate` alias, and that `_deriving` is filtered out of the graph by its underscore, so it is deliberately not registered.

Position behaviour is compile-probed on the merged branch: 9 legitimate spellings compile (including `todo.amount.sum().cast(to: .text)` staying inside the embed, and `todo.name` reaching the `name` column rather than the embed name), and 5 mispositioned ones are rejected — among them `cast(to:).jsonText(_:).eq(_:)`, the `cost::text->>k` 400 the flattened design allowed.

PR #1311, opened 2026-09-01
feat(postgrest): add defaultToNull to insert and upsert

grdsdev:
## Why

A bulk payload sends the union of every row's keys as the `columns` parameter, so a column that only *some* rows carry is written as `null` for the rows that omit it — overwriting its database `DEFAULT`. PostgREST's `Prefer: missing=default` asks for the default instead, and Swift had no first-class way to send it. supabase-js exposes this as `defaultToNull`.

`defaultToNull: Bool = true` on the legacy `insert` and `upsert`; `false` appends `missing=default` to `Prefer`.

> [!IMPORTANT]
> **Draft** — blocked on the spec PR that documents this capability: supabase/sdk#130 (awaiting CODEOWNERS). Opening as a draft so CI can answer the ABI question below while that review happens.

## The ABI question this PR exists to answer

[SDK-1613](https://linear.app/supabase/issue/SDK-1613/postgrest-no-defaulttonull-prefer-missingdefault-on-insert-and-upsert) flags that adding even a *defaulted* parameter to a public function can be reported as an `API breakage:` because the mangled name changes — and this is a **shipped** surface (the legacy builder, present in v2.55.1), so it matters.

I could not answer it locally. `scripts/check-for-breaking-api-changes.sh` reports `❌ Breaking API changes detected` on a tree **byte-identical to the baseline** — all 8 modules fail to load with a `libsecp256k1` module redefinition, so every baseline has zero symbols and the non-zero exit is a false positive (`rm -rf .build` first does not help). The gate is green in CI, so it is a local-environment problem, not a broken gate.

**So `api-stability.yml` on this PR is the verdict.** If it flags the parameter, the fix is to swap it for a new overload rather than carry a `!` and a `V3_MIGRATION.md` entry for a purely additive option — say the word and I will reshape it.

## Two deviations from the issue, both deliberate

**1. The header is not gated on the payload shape.** The issue suggested gating it on the existing `[[String: Any]]` array check "for exact parity". `postgrest-js` does the opposite — `if (!defaultToNull) headers.append('Prefer', 'missing=default')` runs unconditionally, and `Array.isArray` gates only the `columns` parameter. Gating would therefore *break* parity, so the header is sent for a single row too. Harmless there: with no `columns` parameter, omitted columns already take their default.

**2. Scope is insert + upsert only.** PostgREST applies `missing` to `PATCH` as well, but supabase-js exposes `defaultToNull` only on these two, so `update` is untouched.

## Unplanned fallout worth a reviewer's eye

Adding the parameter renames both methods in DocC, which broke **nine pre-existing symbol links** across `PostgrestRequestBuilder.swift`, `PostgrestClient.swift` and `Types.swift` — `./scripts/test-docs.sh` failed until each gained `defaultToNull:`. That is most of the non-test diff and the reason it touches four source files instead of one.

I also dropped a cross-link I had added from `insert` to `upsert`: I wrote it against the `PostgrestQueryBuilder` typealias path, which DocC does not resolve (the type is `PostgrestRequestBuilder`). Replaced with a prose mention — a hardcoded symbol path that must be re-edited on every signature change is not worth the link.

## Test plan

New tests in `PostgrestQueryBuilderTests`, each asserting the **exact full** `Prefer` header via the existing curl inline snapshots (a `contains` check cannot catch a *missing* preference):

| Test | `Prefer` |
|---|---|
| `insertDefaultToNullFalseSendsMissingDefault` | `return=minimal,missing=default` |
| `insertDefaultToNullFalseComposesWithReturningAndCount` | `return=minimal,count=estimated,missing=default` |
| `upsertDefaultToNullFalseKeepsEveryOtherPreference` | `existing=value,resolution=merge-duplicates,return=minimal,count=estimated,missing=default` |
| `upsertDefaultToNullFalseComposesWithIgnoreDuplicates` | `resolution=ignore-duplicates,return=representation,missing=default` |

The third is the one that matters most: it proves the new preference composes with a caller-supplied value, `resolution=`, `return=` and `count=` without dropping any — the failure mode that made #1308 (SDK-1626) a silent data bug.

"Option off" is covered by the pre-existing `insert` / `upsert` / `upsertIgnoreDuplicates` snapshots, which still assert their exact `Prefer` values with no `missing=`. That they pass **unchanged** is also the evidence for "existing call sites unchanged".

- [x] `swift test` — 1441 tests in 148 suites passed (12 known issues)
- [x] `./scripts/format.sh`, `./scripts/spell-check.sh`, `./scripts/test-docs.sh` all clean
- [ ] `api-stability.yml` — the open question; see above
- [x] `sdk-compliance.yaml` unchanged — it lists symbols without parameter labels (`PostgrestRequestBuilder.insert`), and no new symbol lands

Fixes SDK-1613

### DIFF
--- a/Sources/PostgREST/Legacy/PostgrestClient.swift
+++ b/Sources/PostgREST/Legacy/PostgrestClient.swift
@@ -119,9 +119,9 @@ public struct PostgrestClient: Sendable {
     /// The `JSONEncoder` used to serialize request bodies.
     ///
     /// Defaults to ``jsonEncoder``, which is pre-configured with Supabase-compatible settings.
-    /// Individual calls to ``PostgrestRequestBuilder/insert(_:returning:count:encoder:)``,
+    /// Individual calls to ``PostgrestRequestBuilder/insert(_:returning:count:defaultToNull:encoder:)``,
     /// ``PostgrestRequestBuilder/update(_:returning:count:encoder:)``, and
-    /// ``PostgrestRequestBuilder/upsert(_:onConflict:returning:count:ignoreDuplicates:encoder:)``
+    /// ``PostgrestRequestBuilder/upsert(_:onConflict:returning:count:ignoreDuplicates:defaultToNull:encoder:)``
     /// can override this per call.
     public let encoder: JSONEncoder
 
@@ -249,9 +249,9 @@ public struct PostgrestClient: Sendable {
   /// Returns a query builder targeting the specified table or view.
   ///
   /// Call ``PostgrestRequestBuilder/select(_:head:count:)`` on the returned builder to begin a
-  /// `SELECT`, or use ``PostgrestRequestBuilder/insert(_:returning:count:encoder:)``,
+  /// `SELECT`, or use ``PostgrestRequestBuilder/insert(_:returning:count:defaultToNull:encoder:)``,
   /// ``PostgrestRequestBuilder/update(_:returning:count:encoder:)``,
-  /// ``PostgrestRequestBuilder/upsert(_:onConflict:returning:count:ignoreDuplicates:encoder:)``, or
+  /// ``PostgrestRequestBuilder/upsert(_:onConflict:returning:count:ignoreDuplicates:defaultToNull:encoder:)``, or
   /// ``PostgrestRequestBuilder/delete(returning:count:)`` for write operations.
   ///
   /// - Parameter table: The name of the table or view to query.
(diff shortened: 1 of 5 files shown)


grdsdev:
## ABI verdict: not breaking ✅

`api-stability` passed on `a9504453`, the commit carrying the signature change:

```
completed  success  a9504453   api-stability.yml
```

So adding a **defaulted** parameter to a public function on a shipped surface does not trip `swift package diagnose-api-breaking-changes`, despite the mangled name changing. That was the open question in the description, and it resolves the clean way:

- no `!` on the title
- no `V3_MIGRATION.md` entry
- no need to swap the parameter for a new overload

Useful precedent for the next option added to these builders.

Remaining `block-merge` failure is expected and self-clearing — the workflow is "Block WIP/Draft Merges" and its log says `❌ This PR is blocked from merging: Draft PR`. It goes green when this is marked ready for review.

Still a draft pending supabase/sdk#130 (the spec), which is the only thing gating this now.

PR #1316, opened 2026-09-07
feat(auth): verify ES256 JWTs locally in getClaims

AndroidPoet:
### Problem

`getClaims` can only verify RS256 signatures on device. `JWTAlgorithm` has a single case, so for a project signed with an ES256 key the `JWTAlgorithm(rawValue:)` lookup returns nil and every call falls back to `GET /user`, even when the JWKS is cached or passed in through `GetClaimsOptions.jwks`. The existing test `getClaims_withProvidedJWKS_shouldStillFallbackForES256` pinned that behaviour with the note "ES256 is not yet supported client-side". That covers more than new hosted projects: a stock local project has no signing key configured, so the CLI falls back to its built-in ES256 key and every local dev setup signs ES256 today. auth-js verifies both RS256 and ES256 locally, so Swift callers on ES256 projects pay a network round trip per call that JavaScript callers do not.

### Fix

Add an `es256` case to `JWTAlgorithm` and a `JWK.p256PublicKey` helper that builds a `P256.Signing.PublicKey` from the JWK's `x` and `y` coordinates. Verification uses swift-crypto, which `Auth` already depends on for PKCE, so it works on Linux as well as Apple platforms. The JWS signature is the raw `r || s` form, which `ECDSASignature(rawRepresentation:)` accepts directly.

Anything that is not a well formed P-256 key (wrong `kty`, wrong `crv`, a coordinate that is not 32 bytes, or a point that is not on the curve) yields no key, and any signature that is not 64 bytes is rejected, so `verify` returns false rather than trapping. `Base64URL.decode` ignores unknown characters rather than failing, so the length check and the on-curve check in `P256.Signing.PublicKey(x963Representation:)` are what guard the key builder. A token whose signature does not verify now throws `jwtVerificationFailed("Invalid JWT signature")` exactly like the RS256 path instead of being forwarded to the server.

Before:

```swift
enum JWTAlgorithm: String {
  case rs256 = "RS256"
}
```

After:

```swift
enum JWTAlgorithm: String {
  case rs256 = "RS256"
  case es256 = "ES256"
}
```

### Tests

`JWTAlgorithmES256Tests` (new file, 10 tests) signs tokens with a throwaway P-256 key generated in the test, so there is no fixture material and no network:

- public key built from JWK matches the signing key
- non-EC key, non P-256 curve, a 31-byte coordinate, and an off-curve point all yield no key
- valid signature verifies
- tampered payload, signature from another key, and a 4 byte signature all fail

`AuthClientTests`:

- `getClaims_withES256JWTAndJWKS_shouldVerifyLocally` registers `/user` as a 401 so any fallback would fail the test, then asserts the claims and header come back from local verification
- `getClaims_withTamperedES256JWT_shouldThrowWithoutFallback` registers `/user` as a 200 so a fallback would succeed, then asserts `jwtVerificationFailed` is thrown

Both new client tests fail on main (the first hits the 401, the second returns claims) and pass with the fix. The old `shouldStillFallbackForES256` test is replaced by these two. The existing no-kid ES256 test still falls back to the server and is unchanged. Full `AuthTests` suite: 320 tests pass.

No public API change. Not verified on Linux locally, but the new code only uses swift-crypto and Foundation.

### DIFF
--- a/Sources/Auth/Internal/JWK+EC.swift
+++ b/Sources/Auth/Internal/JWK+EC.swift
@@ -1,0 +1,20 @@
+import Crypto
+import Foundation
+
+extension JWK {
+  var p256PublicKey: P256.Signing.PublicKey? {
+    guard kty == "EC",
+      crv == "P-256",
+      let x,
+      let xData = Base64URL.decode(x),
+      let y,
+      let yData = Base64URL.decode(y),
+      xData.count == 32,
+      yData.count == 32
+    else {
+      return nil
+    }
+
+    return try? P256.Signing.PublicKey(x963Representation: Data([0x04]) + xData + yData)
+  }
+}
--- a/Sources/Auth/Internal/JWTAlgorithm.swift
+++ b/Sources/Auth/Internal/JWTAlgorithm.swift
@@ -5,10 +5,12 @@
 //  Created by Claude on 06/10/25.
 //
 
+import Crypto
 import Foundation
 
 enum JWTAlgorithm: String {
   case rs256 = "RS256"
+  case es256 = "ES256"
 
   func verify(
     jwt: DecodedJWT,
@@ -29,6 +31,14 @@ enum JWTAlgorithm: String {
       #else
         return false
       #endif
+    case .es256:
+      guard
+        let publicKey = jwk.p256PublicKey,
+        let signature = try? P256.Signing.ECDSASignature(rawRepresentation: jwt.signature)
+      else {
+        return false
+      }
+      return publicKey.isValidSignature(signature, for: message)
     }
   }
 }
--- a/Tests/AuthTests/AuthClientTests.swift
+++ b/Tests/AuthTests/AuthClientTests.swift
@@ -3149,24 +3149,47 @@ extension AuthMockerTests {
     }
 
     @Test
-    func getClaims_withProvidedJWKS_shouldStillFallbackForES256() async throws {
-      // ES256 is not yet supported client-side, so it will fallback to server even with JWKS
-      let jwt =
-        "eyJhbGciOiJFUzI1NiIsImtpZCI6InRlc3Qta2lkIiwidHlwIjoiSldUIn0.eyJzdWIiOiIxMjM0NTY3ODkwIiwiaXNzIjoiaHR0cDovL2xvY2FsaG9zdDo1NDMyMS9hdXRoL3YxIiwiYXVkIjoiYXV0aGVudGljYXRlZCIsImV4cCI6OTk5OTk5OTk5OSwiaWF0IjoxNTE2MjM5MDIyLCJyb2xlIjoiYXV0aGVudGljYXRlZCJ9.dummysignature"
+    func getClaims_withES256JWTAndJWKS_shouldVerifyLocally() async throws {
+      let signer = ES256TestSigner()
+      let jwt = try signer.sign(
+        header: #"{"alg":"ES256","kid":"es256-test-kid","typ":"JWT"}"#,
+        payload:
+          #"{"sub":"1234567890","iss":"http://localhost:54321/auth/v1","aud":"authenticated","exp":9999999999,"iat":1516239022,"role":"authenticated"}"#
+      )
 
-      // JWK is Codable, no custom init needed
-      let jwkDict: [String: Any] = [
-        "kty": "EC",
-        "kid": "test-kid",
-        "alg": "ES256",
-        "crv": "P-256",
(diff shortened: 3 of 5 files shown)


grdsdev:
Reviewed this against a live local stack (GoTrue **v2.196.0**, CLI 2.115.0). The fix does what it claims — thanks for it. I've pushed three integration tests to this branch (e016e57) covering the gaps below.

## Verified live

The live `.well-known/jwks.json`:

```json
{"alg":"ES256","crv":"P-256","kty":"EC","use":"sig","key_ops":["verify"],
 "kid":"b81269f1-21d8-4f2e-b719-c2240a840d90",
 "x":"M5Sjqn5zwC9Kl1zVfUUGvv9boQjCGd45G8sdopBExB4",
 "y":"P6IXMvA2WYXSHSOMTBH2jsw_9rrzGy89FjPf6oOsIxQ"}
```

Driving `getClaims` with a token that GoTrue actually signed (`{"alg":"ES256","kid":"b81269f1-…"}`) plus that JWKS:

| | `main` (32045417) | this PR |
|---|---|---|
| valid token | `getClaims` → `AuthClient.user(jwt:)` → `HTTPClient.send` — a **`GET /user` round trip**, confirmed by backtrace, even with `jwks` passed in options | claims returned in **~2 ms, zero network** |
| payload tampered to `role: service_role` | forwarded to the server | `jwtVerificationFailed("Invalid JWT signature")` |

## Findings

**1. The premise is understated.** The description says ES256 is "the default for new projects." It's broader than that: the CLI resolves signing keys as `on(...) ?? [HY]`, where `HY` is a hardcoded ES256 key — and the `kid` in the live JWKS above *is* that key. So every stock local dev project already signs ES256 today, and `getClaims()` pays a `GET /user` on every call. Worth saying, since it widens who this fixes.

**2. The load-bearing assumption isn't tested.** `getClaims` picks the verifier from `signingKey.alg` — the **JWK's** field, not the JWT header's. Every test here hand-builds a `JWK` with `alg: "ES256"` hardcoded, so none would catch a server that omitted it. It does hold (live JWKS above; GoTrue's `internal/conf/jwk.go` builds the public key via `jwk.PublicKeyOf`, which copies `alg`), but nothing guarded it. `wellKnownJWKSServesES256Key` now does.

**3. No foreign signer, so the hash choice is unverified.** `ES256TestSigner` signs with swift-crypto and `JWTAlgorithm.verify` checks with swift-crypto, both hashing SHA-256 through the same `DataProtocol` overload. That loop passes even if the hash were wrong for RFC 7518. I confirmed it two independent ways — an openssl-signed token using the live key, and the GoTrue-signed token above — and `getClaimsWithLiveES256Session` keeps it covered.

**Minor, not a bug.** The description says "bad base64url … yields no key." `Base64URL.decode` uses `.ignoreUnknownCharacters`, so it doesn't actually reject garbage — measured: `"@@@@"` → 0 bytes and `"!!!!invalid-base64!!!!"` → 10 bytes, both non-`nil`. What protects `p256PublicKey` is the `count == 32` check plus `P256.Signing.PublicKey(x963Representation:)` rejecting off-curve points. So the code is safe, but `p256PublicKeyRejectsMalformedCoordinates` passes on the length check rather than the decode failure it names.

**Checked and fine:** no algorithm confusion, since the verifier comes from the JWK rather than the token header — a token claiming `alg: RS256` against an EC JWK gets the ES256 verifier and fails. `verify` also signs over the raw `jwt.raw.header`/`jwt.raw.payload` segments, so there's no re-encoding mismatch. 1263 unit tests pass.

## About the tests I pushed

They run against the existing project — no new config or fixture, since the stock CLI default is already ES256. They use `signUp` directly rather than `signUpIfNeededOrSignIn` because `mockEmail()` is random per call, so the leading `signIn` attempt is a wasted round trip.

One caveat: only `wellKnownJWKSServesES256Key` ran on my machine. The two session-based tests are blocked locally by a **pre-existing** crash unrelated to this PR — every Auth integration test segfaults on macOS as soon as the SDK makes an HTTP request (`EXC_BAD_ACCESS` at `swift_task_dealloc`, through the default transport closure in `AuthClient.Configuration.init` and `RetryRequestInterceptor.retry`). It reproduces at 32045417 with `multipleAuthInstances`, which this PR doesn't touch. CI is unaffected because the integration job runs on Linux, so all three should run there. Filing that separately.

AndroidPoet:
Thanks for the live verification and the integration tests. Addressed the two points: the malformed-coordinates test now asserts what actually guards the key builder, split into a 31-byte coordinate case and an off-curve point case (773906de), and the description now states the local dev premise and the base64url behaviour accurately.

grdsdev (approved):

PR #1317, opened 2026-09-07
fix(postgrest): honor the relation schema in typed queries

AndroidPoet:
### Problem

`@Table` accepts a `schema:` argument and the macro emits it as `static let schema`, but nothing reads it. `PostgrestClient.from(_ relation:)` builds the request from `R.relationName` only, so a relation declared as `@Table("secrets", schema: "private")` is queried in `public`. There is no compiler or runtime signal; the request simply goes to the wrong schema, and the only test touching the value asserts `Todo.schema == "public"` without issuing a request.

Every typed entry point routes through this one call, so reads, inserts, updates, upserts and deletes were all affected.

### Fix

`from(_:)` now scopes the client to the relation's declared schema before building the request, which sends `Accept-Profile` on reads and `Content-Profile` on writes through the existing header logic.

Two rules keep this from surprising anyone:

- A client already scoped with `schema(_:)` keeps that schema. The migration guide documents `client.schema("public").from(Todo.self)` as the typed spelling, and per-tenant schemas rely on the call site winning over a type-level default.
- A relation that declares `public` on an unscoped client sends no profile header, which is what happens today. The macro defaults `schema` to `"public"`, so without this rule every existing typed request would gain a header it never had.

Before:

```swift
public func from<R: PostgrestRelation>(_ relation: R.Type) -> PostgrestTypedSource<R> {
  PostgrestTypedSource(builder: from(R.relationName))
}
```

After:

```swift
public func from<R: PostgrestRelation>(_ relation: R.Type) -> PostgrestTypedSource<R> {
  PostgrestTypedSource(builder: scoped(to: relation).from(R.relationName))
}

private func scoped<R: PostgrestRelation>(to relation: R.Type) -> PostgrestClient {
  guard configuration.schema == nil, R.schema != "public" else { return self }
  return schema(R.schema)
}
```

If you would rather the relation always win, or always send the header even for `public`, both are one-line changes and I am happy to switch.

### Tests

Four new tests in `PostgrestTypedSourceTests`, using a hand-written `Secret` relation in schema `private`:

- a read sends `Accept-Profile: private` and no `Content-Profile`
- an insert sends `Content-Profile: private` and no `Accept-Profile`
- `client.schema("tenant_a").from(Secret.self)` sends `Accept-Profile: tenant_a`
- the existing `public` relation on an unscoped client sends neither header

The two `Secret` tests fail on main (no profile header is sent) and pass with the fix. Full `PostgRESTTests` and `PostgrestMacrosTests` suites pass.

No public API change.

### DIFF
--- a/Sources/PostgREST/Query/PostgrestTypedSource.swift
+++ b/Sources/PostgREST/Query/PostgrestTypedSource.swift
@@ -13,11 +13,6 @@ extension PostgrestClient {
   /// let todos = try await client.from(Todo.self).select().execute().value
   /// ```
   ///
-  /// The request targets the schema the relation declares through
-  /// ``PostgrestRelation/schema``, unless this client was already scoped with ``schema(_:)``, in
-  /// which case the client's schema wins. A relation in `public` on an unscoped client sends no
-  /// profile header, which is PostgREST's default.
-  ///
   /// - Parameter relation: The relation type to query.
   /// - Returns: A ``PostgrestTypedSource`` for that relation.
   public func from<R: PostgrestRelation>(_ relation: R.Type) -> PostgrestTypedSource<R> {
--- a/Sources/PostgREST/Query/PostgrestTypedSource.swift
+++ b/Sources/PostgREST/Query/PostgrestTypedSource.swift
@@ -13,10 +13,20 @@ extension PostgrestClient {
   /// let todos = try await client.from(Todo.self).select().execute().value
   /// ```
   ///
+  /// The request targets the schema the relation declares through
+  /// ``PostgrestRelation/schema``, unless this client was already scoped with ``schema(_:)``, in
+  /// which case the client's schema wins. A relation in `public` on an unscoped client sends no
+  /// profile header, which is PostgREST's default.
+  ///
   /// - Parameter relation: The relation type to query.
   /// - Returns: A ``PostgrestTypedSource`` for that relation.
   public func from<R: PostgrestRelation>(_ relation: R.Type) -> PostgrestTypedSource<R> {
-    PostgrestTypedSource(builder: from(R.relationName))
+    PostgrestTypedSource(builder: scoped(to: relation).from(R.relationName))
+  }
+
+  private func scoped<R: PostgrestRelation>(to relation: R.Type) -> PostgrestClient {
+    guard configuration.schema == nil, R.schema != "public" else { return self }
+    return schema(R.schema)
   }
 }
 
--- a/Tests/PostgRESTTests/PostgrestTypedSourceTests.swift
+++ b/Tests/PostgRESTTests/PostgrestTypedSourceTests.swift
@@ -28,6 +28,26 @@ struct PostgrestTypedSourceTests {
     static let columns = Columns()
   }
 
+  struct Secret: PostgrestWritableRelation {
+    static let relationName = "secrets"
+    static let schema = "private"
+    static let selectString = "*"
+
+    var id: Int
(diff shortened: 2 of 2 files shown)


grdsdev:
Thanks @AndroidPoet for this contribution, but I plan to solve this issue in a different way, my plan is to also have a typed/generated schema types, for example:

```swift
enum PublicSchema: Schema {}
enum PrivateSchema: Schema {}

@Table("todos") // defaults to PublicSchema when omitted.
struct Todo {
    ///...
}

@Table("secret", schema: PrivateSchema.self)
struct Secrets {
    ///...
}

try client.schema(PrivateSchema.self).from(Todo.self) // compiler error as Todo isn't on PrivateSchema
try client.schema(PrivateSchema.self).from(Secrets.self) // valid
```

AndroidPoet:
Makes sense, the typed schema approach is clearly better since it catches the mismatch at compile time. How would you like to handle this one: should I close it, or would it help if I took a pass at the Schema type design you sketched so you can review rather than write it? Happy either way, and no problem if you would rather do it yourself.

AndroidPoet:
Following up on the typed schema plan, since this PR is blocked on it.

Sketching it against the current typed layer, it lands on two patterns that are already in `PostgREST`, which is what makes me think it is a small change rather than new machinery:

```swift
public protocol PostgrestSchema { static var name: String { get } }
public enum PublicSchema: PostgrestSchema { public static let name = "public" }
```

Those markers are the same shape as the phases in `PostgrestRequestBuilder.swift`, uninhabited enums used as a compile-time-only parameter. The mismatch you sketched then falls out of the same technique `select(_:)` already uses with `where S.Source == R`:

```swift
extension PostgrestClient {
  public func schema<S: PostgrestSchema>(_ schema: S.Type) -> PostgrestSchemaScope<S>
}

public struct PostgrestSchemaScope<S: PostgrestSchema>: Sendable {
  public func from<R: PostgrestRelation>(_ relation: R.Type) -> PostgrestTypedSource<R>
  where R.Schema == S
}
```

On the relation side, `static var schema: String` becomes `associatedtype Schema: PostgrestSchema = PublicSchema`. The default is what keeps hand-written conformances and the existing test fixtures compiling without declaring anything, and `@Table("secret", schema: PrivateSchema.self)` emits `typealias Schema = PrivateSchema` where it emits `static let schema = "..."` today. `TableMacro` reads that argument as a string literal right now, so that part is a few lines.

`from(_:)` can then read the schema off the relation type, which is the bug this PR was opened for.

Three calls that are yours to make:

- naming: `PostgrestSchema` to match `PostgrestRelation` and `PostgrestSelection`, or plain `Schema` as you wrote it.
- whether `schema(_ schema: String)` stays untouched for the untyped path. I assume yes.
- `PostgrestRelation.schema` is registered in `sdk-compliance.yaml`, so dropping it is a registered-symbol removal the matrix check will flag until the entry moves to the associated type.

Do you want me to implement this so you have something to review, or would you rather write it yourself? I am happy either way, and happy to close this PR in favour of it.

grdsdev:
That is great plan and aligns with what I have in mind, feel free to implement it that way.

Name it `PostrestSchema` rather than `Schema`. And keep the string fallback.

AndroidPoet:
Superseded by #1372, which implements the typed shape agreed here: the relation names its schema with a type, `from(_:)` reads it, and `client.schema(PrivateSchema.self)` rejects a relation from another schema at compile time. The string fallback stays, as asked.

Closing this one so the review lives in a single place.

PR #1333, opened 2026-09-14
feat(helpers): add SupabaseError root and HTTPErrorResponse

grdsdev:
Part 1 of 6 of the error revamp (SDK-1788). Stacked: this PR is the root; #1334, #1335, #1336, #1337 and #1338 build on it.

## What

Adds two additive public types to `Helpers`, re-exported by every module:

- `SupabaseError`: the root protocol every module error will conform to. Three requirements: `message`, `response: HTTPErrorResponse?`, `underlyingError`. Default `errorDescription` returns `message`.
- `HTTPErrorResponse`: status code, `HTTPFields` headers, raw body, and `requestID` read from the `sb-request-id` header Supabase attaches to every response.

Nothing conforms yet. No breaking change in this PR.

## Why

Six public error types shared no root, none carried headers or a request id, and Realtime's error was not even public. A caller could not write one `catch` for the SDK. Later PRs in this stack make each module error a struct with an open `Kind` that conforms here.

## Testing

`HTTPErrorResponseTests` and `SupabaseErrorTests` cover `requestID`, the package initializer from the internal HTTP response, hashability, and the shared `description` layout.

Refs SDK-1788. Design: the stack's migration sections land in `V3_MIGRATION.md` with each module PR.


spydon (approved):
Lgtm
TIL that it's idiomatic for Swift code to use extensions in the way that you did here, one wouldn't do that with Dart extensions.

PR #1335, opened 2026-09-14
feat(storage)!: make StorageError a struct with kind, server payload and HTTP context

grdsdev:
Part 3 of 6 of the error revamp (SDK-1788). Stacked on #1334.

## What

`StorageError` becomes a struct conforming to `SupabaseError`, with `kind` (`.server`, `.unexpectedResponse`, `.transport`, `.decoding`, `.invalidURL`), `serverError`, `response` and `underlyingError`. The body Storage returns is decoded into the nested `StorageError.ServerError` (wire fields `statusCode`, `error`, `message`), present exactly when `kind == .server`. `StorageError` itself is no longer `Decodable`.

Storage no longer throws `HTTPError` for an unrecognized body (that is `.unexpectedResponse` with the raw body in `response`), and the six `URLError(.badURL)` throws in URL-building helpers become `.invalidURL`. Decoding failures on success bodies are wrapped as `.decoding` through a same-module shadow of `decoded(as:decoder:)`, so no call site changed.

## Breaking

`statusCode` and `error` no longer exist on `StorageError`; use `response?.statusCode` and `serverError?.error`. `JSONDecoder().decode(StorageError.self, ...)` no longer compiles. See `V3_MIGRATION.md`, section "`StorageError` gains `kind` and `response`; `statusCode`, `error` and the decoded `message` move to `serverError`".

## Testing

Unit tests per kind plus cancellation and custom-fetch passthrough. The five integration snapshots that dumped the old struct are now field asserts.

Refs SDK-1788.


Sources/Storage/StorageError.swift
```diff
@@ -1,44 +1,87 @@
...
+    /// A non-2xx status whose body was not a Storage error payload. ``StorageError/response``
+    /// has the raw body.
+    public static let unexpectedResponse: Kind = "unexpectedResponse"
+    /// The request never completed. ``StorageError/underlyingError`` is usually a `URLError`.
+    public static let transport: Kind = "transport"
+    /// A success body could not be decoded. ``StorageError/underlyingError`` is usually a
+    /// `DecodingError`.
+    public static let decoding: Kind = "decoding"
+    /// A URL could not be built from the configuration and the given path. No request was sent.
+    public static let invalidURL: Kind = "invalidURL"
```
spydon, line 41: Why a RawRepresentable struct with static members instead of enum Kind: String?

spydon: Ah, to avoid future breaking changes I guess.

spydon (approved):

PR #1337, opened 2026-09-14
feat(auth)!: make AuthError a struct with kind, HTTP context and wrapped failures

grdsdev:
Part 5 of 6 of the error revamp (SDK-1788). Stacked on #1336.

## What

`AuthError` becomes a struct conforming to `SupabaseError`: `kind`, `message`, `errorCode`, `weakPasswordReasons`, `response`, `underlyingError`. `Kind` mirrors the old cases (`.api`, `.sessionMissing`, `.weakPassword`, `.pkceGrantCodeExchange`, `.implicitGrantRedirect`, `.jwtVerificationFailed`) and adds `.webAuthn`, `.unexpectedResponse`, `.transport`, `.decoding`. `AuthError.sessionMissing` stays as a static value, so `throw AuthError.sessionMissing` compiles unchanged. `HTTPURLResponse` leaves the public error surface. The internal `WebAuthnError`, which could leak from passkey flows, is folded into `.webAuthn`.

This PR also carries the branch-wide correction that `.transport` wraps only `URLError`: an error thrown by a custom `fetch` or by the `accessToken` closure (which `SupabaseClient` resolves inside its shared fetch adapter) now propagates as itself in Functions, Storage, PostgREST and Auth. A Supabase umbrella test covers that contract.

## Breaking

Every `case`-based pattern and the `~=` operator stop compiling. `AuthError` is not `Equatable`. PKCE errors carry `error_code` in `errorCode` and the OAuth `error` as a message prefix. See `V3_MIGRATION.md`, section "`AuthError` is now a struct, not an enum".

`signOut` still swallows 401, 403 and 404 from `/logout` whether or not the body decoded; no behavior change there.

## Testing

`AuthErrorTests` rewritten for the struct; every enum-pattern test in `AuthTests` and the integration tests now asserts `kind`, `errorCode`, `message`, `response`. New tests for transport wrapping, unexpected bodies, decoding wrapping, cancellation and custom-fetch passthrough, and the `.webAuthn` parser paths.

Refs SDK-1788, SDK-1472.


spydon (approved):
Lgtm
1337 😎

PR #1342, opened 2026-09-15
perf(helpers)!: stream request bodies instead of buffering them before upload

grdsdev:
`URLSessionTransport` collected every `.stream` request body into memory, with no cap, before starting the request. Request bodies now stream.

**Apple platforms.** A `.stream` body goes through `uploadTask(withStreamedRequest:)`. `StreamingTaskDelegate` answers `needNewBodyStream` by handing `URLSession` the read end of a bound `InputStream`/`OutputStream` pair; the new `HTTPBodyOutputStreamBridge` pulls one chunk at a time and writes only while the stream has space, so memory stays at the pair's buffer plus one chunk. `.file` bodies go through `uploadTask(with:fromFile:)` and `.data` bodies through `dataTask`, all on the same delegate. A `.known(n)` length is sent as `Content-Length` and the body must yield exactly `n` bytes, otherwise the request fails with the new `HTTPBodyLengthMismatchError` instead of stalling until the timeout (too few) or truncating server-side (too many). `.unknown` omits `Content-Length` so URLSession uses chunked transfer.

**One-shot bodies and redirects.** A 307/308 keeps the method and resends the body. For a `.single` body the transport does not follow it and returns the redirect response as-is (the `net/http` rule), rather than failing halfway through. If `URLSession` asks for the body again for any other reason, such as an authentication retry, the request fails with `HTTPBodyAlreadyConsumedError`. `.multiple` bodies replay.

**Linux.** swift-corelibs-foundation has no per-task delegates, so a `.stream` body is spooled to a temporary file, checked against a `.known` length, and uploaded from there. Memory stays flat and redirects are followed for every body kind because the spooled copy can be resent.

**`HTTPBody`.** `init(_ chunks:)`, `init(fileURL:)` and `reportingProgress` are pull-based (`AsyncThrowingStream(unfolding:)`, one `next()` per pull) instead of pumping the source into an unbounded stream from a detached `Task`; that pump kept the streamed upload at two copies even after the transport change. `reportingProgress` keeps the receiver's storage and installs exactly one reporting path: a `.stream` body reports as chunks are pulled, a `.file` or `.data` body still uploads flat and reports `URLSession`'s sent-byte counts through the package-level `onUploadProgress` hook, so a middleware that buffers the body (`LoggerInterceptor` keeps the hook on the body it re-wraps) can never double-count.

**Breaking change** (`V3_MIGRATION.md`): on Apple platforms a `.single` request body that receives a 307/308 now gets the redirect response back instead of having it followed, and a `.known` streamed body whose byte count does not match fails with `HTTPBodyLengthMismatchError`.

## Test plan

- `swift test` on the rebased branch: 1515 tests, 150 suites, pass (1 pre-existing known issue). The bound-pair read helpers in the tests run on a global queue, because a blocking bound-pair read spins the calling thread's run loop and Swift Testing can place a test on the main thread in a full parallel run.
- `HTTPBodyOutputStreamBridgeTests`: back-pressure with a 150-byte buffer (pull count stays at 2 until the reader drains), `.single` replay fails with `HTTPBodyAlreadyConsumedError`, `.multiple` replays, body shorter and longer than the declared length fail with `HTTPBodyLengthMismatchError`.
- `URLSessionTransportTests`: streamed body pulled while the request is in flight (Apple) / collected first (Linux), `didSendBodyData` forwarded for file bodies and not for streamed ones, second body stream on a `.single` body cancels the task and surfaces `HTTPBodyAlreadyConsumedError`, 307/308 not followed for `.single` bodies.
- `HTTPBodyTests`: pull-on-demand, `reportingProgress` keeps storage and reports once per path. `LoggerInterceptorTests`: progress hook survives the re-wrap.
- `RetryRequestInterceptorTests` unchanged and passing.
- Memory probe `upload-transport-stream`, 200 MiB body, local Supabase: +377.5 MB before, +5.9 MB after. `upload-transport-file` +4.5 to +7.6 MB across runs (noise band), `download-transport-iterate` +0.9 MB. The probe (`MemoryProbeIntegrationTests`, opt-in via `INTEGRATION_TESTS=1 MEMORY_PROBE=1`) is included so the numbers are reproducible.
- `swift-format lint --strict`, cspell and `./scripts/test-docs.sh` clean.

Fixes SDK-1842


Sources/Helpers/HTTP/URLSessionTransport.swift
```diff
@@ -58,47 +71,67 @@ public struct URLSessionTransport: ClientTransport {
...
+        try await body?.write(to: url)
+        fileURL = url
+      case .data?, nil:
+        break
       }
-    }
-    return try await streamResponse(for: urlRequest)
+      let (data, response) =
+        if let fileURL {
+          try await session.upload(for: urlRequest, fromFile: fileURL)
```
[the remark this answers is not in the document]
grdsdev: Not applying this one. On Linux the streamed body is spooled to a temporary file before the upload starts, so swift-corelibs-foundation can resend it on a 307/308 and following the redirect gives the caller the 2xx it wanted. Refusing there would make Linux strictly worse to mirror a limitation that only exists on Apple platforms, where the body cannot be produced twice.

What was wrong is the doc: it stated the rule without saying it is Apple-only. Fixed that in the transport doc and the V3 migration entry. The other finding (checking the spooled size against the declared length) is valid and is fixed in the same commit.

Sources/Helpers/HTTP/HTTPBody.swift
```diff
@@ -227,35 +233,55 @@ extension HTTPBody {
...
+
+/// Mutable state for a pull closure. The consuming `AsyncThrowingStream` serializes pulls, so
+/// the value is never touched from two places at once.
+private final class Box<Value>: @unchecked Sendable {
+  var value: Value
+
+  init(_ value: Value) {
+    self.value = value
+  }
+}
```
spydon, line 287: This looked a but weird I think, but discussed it with the clanker and it's probably better than the options.

spydon (approved):

Sources/Helpers/HTTP/HTTPBody.swift
grdsdev: This is a well known Swift trick for holding a value type as reference type.

This was even added as a type in the standard library, but unfortunately available only on Swift 6.4 https://github.com/swiftlang/swift-evolution/blob/main/proposals/0517-uniquebox.md

PR #1344, opened 2026-09-15
docs: spell out that a cancelled request surfaces as .transport, not CancellationError

grdsdev:
Stacked on #1341 — review and merge that first.

`URLSession`'s async APIs report a cancelled `Task` as `URLError(.cancelled)`,
not `CancellationError`. Every module wraps a `URLError` as its module error
with kind `.transport`, so a cancelled request surfaces as
`StorageError(transport)`, `PostgrestError(transport)`, and so on. The V3
migration guide only promised that `CancellationError` propagates as itself,
which reads as "cancelling gives you a `CancellationError`" — it does not, so a
caller guarding an error banner with `error is CancellationError` shows the
banner for a request the user cancelled.

The decision here is to **keep the wrapping** and document it. Rethrowing the
`URLError` unchanged and converting it to `CancellationError` were both
considered and rejected: the latter misreports a `URLError(.cancelled)` that
did not come from `Task` cancellation (`URLSession.invalidateAndCancel()`) and
loses the original error.

So this is a docs change plus one test:

- `V3_MIGRATION.md`, in the existing "Network and decoding failures are wrapped
  in the module error" section: a before/after showing the `underlyingError`
  check, and an explicit note that the break compiles silently.
- `cancelledURLErrorIsWrapped()` in `Tests/StorageTests/StorageFileAPITests.swift`,
  placed directly after `cancellationIsNotWrapped()` so the two forms are
  contrasted where the confusion is. One test, not one per module: each module
  already has a `transportFailureIsWrapped` test proving `URLError` →
  `.transport`, and `.cancelled` is deliberately not special.

SDK-1849 also asked for `shouldRetry` to stop retrying a cancelled request.
#1341 already does that — it deletes PostgREST's legacy `shouldRetry` and routes
every module through `RetryRequestInterceptor`, which returns
`error.code != .cancelled` and is covered by `doesNotRetryCancelledURLError()`.
Nothing was needed here.

## Testing

Per-target, all green:

| Target | Result |
| --- | --- |
| AuthTests | 329 passed |
| StorageTests | 193 passed |
| PostgRESTTests | 282 passed (1 known issue) |
| FunctionsTests | 49 passed |
| RealtimeTests | 276 passed |
| HelpersTests | 154 passed |
| SupabaseTests | 28 passed |

`swift-format lint --recursive --strict Sources Tests` and
`./scripts/spell-check.sh` both clean.

## Risks

Documentation and one test; no source change, so no runtime risk.

Unrelated, but worth a separate look: whole-package `swift test` segfaults on
this base (signal 11, before any test completes). It reproduces with this
commit reverted, so it predates the branch — but CI only runs the per-platform
jobs, which pass, so nothing currently catches it.

Fixes SDK-1849


V3_MIGRATION.md
```diff
@@ -2276,6 +2276,28 @@ Errors thrown by your own code that runs inside the request — a custom `Client
 error is in `underlyingError`. `CancellationError` is never wrapped and still propagates as
 itself.
 
+Cancelling a request is the case worth spelling out. `URLSession`'s async APIs do not throw
+`CancellationError` when the enclosing `Task` is cancelled — they throw `URLError(.cancelled)`,
```
spydon, line 2283: `URLError` sounds a bit strange to use for this use-case?

grdsdev: `URLError` is the error thrown by Apple's networking stack, for some reason they thrown a `URLError.cancelled` instead of `CancellationError` which is the error thrown by Swift structured concurrency, it is weird.

That part states the decision I took of not wrapping `CancellationError`, but do wrap the `URLError(.cancelled)` even if they mean the same thing in practice, since Apple's own framework throws it differently, I decided to also treat it differently.

spydon: Alright, Apple being weird then 😄

spydon (approved):

PR #1358, opened 2026-09-16
feat(storage): add CDN cache purging for files and buckets

AndroidPoet:
Adds `purgeCache(path:transformationsOnly:)` on a bucket's file API and `purgeCache(bucket:transformationsOnly:)` on the Storage client, over Storage's `DELETE /cdn/{bucket}/{path}` and `DELETE /cdn/{bucket}` endpoints. After replacing a file at the same path, a backend can drop the stale CDN copy instead of waiting out `cacheControl` or adding a cache-busting parameter to every URL.

```swift
try await supabase.storage.from("avatars").purgeCache(path: "user123.png")
try await supabase.storage.purgeCache(bucket: "avatars", transformationsOnly: true)
```

Marks `storage.file_buckets.purge_cache` and `storage.file_buckets.purge_bucket_cache` implemented in `sdk-compliance.yaml`. The JS and Flutter clients already ship both.

## Shape

- Both return nothing, like `emptyBucket` and `deleteBucket`. Storage answers `{"message":"success"}`, which carries nothing a caller can act on.
- `transformationsOnly:` becomes `?transformations=true`, sent only when `true`. JS and Flutter name it `transformations`, but `transformations: true` reads as "purge the variants too", when it actually purges only the variants and leaves the original cached.
- The file path goes through the same `_getFinalPath` and `appendingPathComponent` as `info` and `exists`, so a leading `/` is stripped and the key is percent-encoded. The JS client first shipped this without encoding and fixed it afterwards (supabase/supabase-js#2545); `purgeCachePercentEncodesThePath` pins it here.
- The bucket id is interpolated as-is, like every other bucket route in this client.

## Worth knowing

- Storage serves these routes only to the service role, and only when a CDN purge endpoint is configured (`CDN_PURGE_ENDPOINT_URL`, plus the `purgeCache` tenant feature on multi-tenant deployments). The doc comments say so.
- The bucket route and the `transformations` flag arrived in Storage v1.61.0 (supabase/storage#1133). On an older self-hosted Storage, `purgeCache(bucket:)` gets a route-not-found 404, and `purgeCache(path:transformationsOnly: true)` purges the whole file, because the older file route ignores the query.

## Testing

Five request-snapshot tests: a file, a file with `transformationsOnly`, a file path that needs percent-encoding, a bucket, and a bucket with `transformationsOnly`. Dropping the query item makes both `transformationsOnly` tests fail.

I also ran all four calls through the SDK with a throwaway test (not part of this PR) against Storage v1.77.4 on the local stack, with `CDN_PURGE_ENDPOINT_URL` pointed at a stub that logs what Storage forwards. That run predates the rename to `purgeCache(bucket:)`, which only changes the call site: the request snapshots are the same.

```
POST /purge {"type":"object","tenant":{"ref":"stub"},"bucketId":"avatars","objectName":"folder/my file.png"}
POST /purge {"type":"object-transforms","tenant":{"ref":"stub"},"bucketId":"avatars","objectName":"folder/my file.png"}
POST /purge {"type":"bucket","tenant":{"ref":"stub"},"bucketId":"avatars"}
POST /purge {"type":"bucket-transforms","tenant":{"ref":"stub"},"bucketId":"avatars"}
```

The two file calls passed `folder/my file.png` and `/folder/my file.png`, and both reached the purge endpoint as the same object name. The bucket call with the anon key threw `StorageError` with kind `.server`, status 403 and `Access denied: Invalid role`.

The capability matrix checks from `validate-sdk-compliance-swift.yml` (pinned `capability-matrix-v1.6.0`) pass locally against `main`, with symbol graphs dumped and normalized the way CI does it.

```
$ swift test --skip IntegrationTests   # exit 0
9 test runs, all passed: 1416 tests in 141 suites, 174 of them in StorageTests

$ swift-format lint --recursive --strict Sources Tests   # exit 0
$ ./scripts/spell-check.sh   # exit 0, 413 files checked, 0 issues
$ ./scripts/test-docs.sh   # its docbuild and warning filter: exit 0, no warnings
```

## Review first

The `transformationsOnly:` label, which departs from the JS and Flutter spelling on purpose.


grdsdev (changes requested):
Per the Swift API Design Guidelines (clarity at the point of use / include all words needed to avoid ambiguity), the two new purge methods label their identifying argument inconsistently:

```swift
try await supabase.storage.from("avatars").purgeCache(path: "user123.png")
try await supabase.storage.purgeBucketCache("avatars", transformationsOnly: true)
```

`purgeCache(path:)` labels its string because an unlabeled string could mean anything (path, key, URL). `purgeBucketCache(_:)` does the same job — identifying the thing whose cache gets purged — but leaves it unlabeled. Read on its own, `purgeBucketCache("avatars", ...)` doesn't make it clear "avatars" is a bucket id, especially with `Cache` as the noun immediately preceding the argument slot (a reader could momentarily read it as identifying which cache, not which bucket).

Suggest labeling it to match the file-scoped method, and dropping "Bucket" from the base name since the label now carries that meaning (no collision — different receiving types):

```swift
public func purgeCache(bucket: String, transformationsOnly: Bool = false) async throws
```

```swift
storage.from("avatars").purgeCache(path: "user123.png")
storage.purgeCache(bucket: "avatars", transformationsOnly: true)
```

Both call sites now read the same way and disambiguate their argument the same way.

AndroidPoet:
Renamed to `purgeCache(bucket:transformationsOnly:)` in 97f5ee11, along with its parameter doc, the `SupabaseStorageClient` topics entry, the two bucket tests and the `sdk-compliance.yaml` symbol. I also merged `main` so it builds on top of #1355. The description now uses the new name.

grdsdev (approved):

PR #1367, opened 2026-09-17
refactor: prune redundant and history-narrating comments

grdsdev:
## What

An audit of every non-header `//` comment in `Sources` (653 lines) against the repo's comment rules: names and structure carry the "what", and only an outside-the-code "why" or genuinely dense syntax earns a comment.

Net: 47 comment lines removed across 14 files. No behavior change.

## Deleted — restated the line below them

- `Auth/AuthClient.swift` — 11, nearly all in `fetchJWK` and `getClaims` (`// Fetch from well-known endpoint`, `// Cache the JWKS globally`, `// Decode claims and header`, …)
- `RealtimeV2/WebSocket/URLSessionWebSocket.swift` — 4, including `// Network protocol error.` on the `EPROTO` case and a `// Update state when connection closes` that duplicated the comment three lines above it
- `RealtimeV2/ConnectionManager.swift`, `RealtimeChannelV2.swift`, `PushV2.swift`, `RealtimeSerializer.swift` — 6
- `Auth/Types.swift` — the `// RSA-specific fields` / `// EC-specific fields` / `// Symmetric key field` labels on `JWK`, where the `///` docs immediately below already say it
- `Auth/Internal/APIClient.swift`, `Auth/Internal/SessionStorage.swift` — 2
- `TestHelpers/URLRequestSnapshot.swift` — the `// HTTP Method` / `// Headers` / `// Body` / `// Cookies` / `// URL` section labels

## Trimmed — kept the "why", dropped the history

`StoredProperty.swift`, `PostgrestUpdate.swift`, `WaitUntil.swift` and `RealtimeChannelV2.swift:1084` each explained a real hazard and then narrated the old code that hit it. Each now states the hazard without the "previously".

The `!isTesting` comment in `broadcast(event:message:)` was two lines of first-person rambling; it is now two lines of why.

The two `TableMacro.swift` comment essays (6 and 15 lines) are now 4 and 9, keeping the `postgres-meta` optionality rule and the reason there is no `Update` shape.

## Note for review

`TestHelpers/URLRequestSnapshot.swift` is vendored from swift-snapshot-testing. Removing those five labels widens the diff against upstream — say the word and I will restore them.

## Verification

- `./scripts/format.sh`
- `xcrun swift-format lint --recursive --strict Sources Tests` — passes
- `swift build` — succeeds


spydon (approved):
It's so annoying that all LLM models keep on doing this even with strict instructions that they shouldn't. 😅

PR #1370, opened 2026-09-17
feat(storage): expose the error code from Storage error responses

AndroidPoet:
Adds `code` to `StorageError.ServerError`, decoded from the `code` field of Storage's error bodies. It is the machine-readable identifier listed on the [Storage error codes](https://supabase.com/docs/guides/storage/debugging/error-codes) page, so a caller can branch on it instead of matching `error` or the message.

```swift
do {
  try await supabase.storage.from("avatars").upload(path: path, data: data)
} catch let error as StorageError where error.serverError?.code == .keyAlreadyExists {
  askToOverwrite()
}
```

Marks `storage.errors.error_codes` implemented in `sdk-compliance.yaml`, in a new Storage errors group, since the sync left it ungrouped.

## Why `error` is not enough

`error` still carries Storage's older identifiers, and one of them covers several failures. From Storage v1.72.1, the version the latest CLI starts:

```
upload an existing object  {"statusCode":"409","error":"Duplicate","message":"The resource already exists","code":"KeyAlreadyExists"}
create an existing bucket  {"statusCode":"409","error":"Duplicate","message":"The resource already exists","code":"BucketAlreadyExists"}
download a missing object  {"statusCode":"404","error":"not_found","message":"Object not found","code":"NoSuchKey"}
```

The first two differ only in `code`. All three arrive with HTTP 400, so the transport status does not help either.

## Shape

- `code: StorageError.Code?`, a `RawRepresentable` struct shaped like `ErrorCode` in Auth: `Decodable, RawRepresentable, Sendable, Hashable`, with `init(rawValue:)` and the shorthand `init(_:)`. The 60 codes Storage defines today are `static let`s on it, each documented.
- An unrecognised code decodes instead of failing, which is why this is a struct and not an enum. `serverErrorKeepsACodeItDoesNotKnow` pins that.
- I nested it as `StorageError.Code` rather than adding a second top-level `ErrorCode`, since `StorageError.Kind` already sits there and a top-level one would collide with Auth's under `import Supabase`. Say the word if you would rather have `StorageErrorCode`.
- It is optional because Storage only sends it from v1.68.0 (supabase/storage#1256). Before that, the error response schema dropped the field. Storage v1.62.5, for example, still answers without it.
- `init` gains a defaulted `code:` between `error:` and `message:`, so existing calls compile unchanged.
- The vector bucket clients throw through the same `StorageApi.execute`, and Storage's vector routes use the same error handler, so vector errors get `code` too.

## Testing

- `serverErrorDecodesTheCode` decodes the missing-object body above into `.noSuchKey`, and `serverErrorDecodesWithOnlyAMessage` now also checks that a body without `code` decodes to `nil`.
- `serverErrorKeepsACodeItDoesNotKnow` decodes `"SomeFutureCode"` and checks it survives as a `Code` with that `rawValue`, so a server newer than the SDK keeps working.
- `nonSuccessStatusCodeExposesTheServerErrorCode` downloads a missing object through the client and checks `kind`, `code`, `error` and the 400 status.
- The integration tests that already assert `error` for a missing bucket, an oversized upload, a disallowed MIME type and a repeated signed upload now assert `code` as well (`NoSuchBucket`, `EntityTooLarge`, `InvalidMimeType`, `KeyAlreadyExists`).

Without the change the tests do not compile (`value of type 'StorageError.ServerError' has no member 'code'`).

I ran the two Storage integration suites locally against Storage v1.72.1, with a proxy for the `/storage/v1` prefix in place of the gateway: 30 tests, all passed. The bodies above come from that same server. That run predates both the rebase onto #1368 and the switch to the typed code, which only rewrote those same expectations (`"NoSuchBucket"` became `.noSuchBucket`). They still compile, but I have not re-run them live since. The unit gates below are from the current tree.

```
$ swift test --skip IntegrationTests   # exit 0
9 test runs, all passed: 1422 tests in 141 suites, 177 of them in StorageTests

$ INTEGRATION_TESTS=1 swift test --filter "StorageClientIntegrationTests|StorageFileIntegrationTests" --no-parallel   # exit 0, 30 tests

$ swift format lint --strict --recursive Sources Tests Package.swift Examples   # exit 0
$ ./scripts/spell-check.sh   # exit 0, 413 files checked, 0 issues
```

The capability matrix checks from `validate-capabilities.yml` (pinned `capability-matrix-v1.6.0`) passed locally before the rebase: validate, drift, and the new-symbol check, which fails against `main`'s `sdk-compliance.yaml` and passes with this one. I dumped the public symbol graph again against a clean `main` build: 64 new symbols, all registered under `storage.errors.error_codes`, and nothing registered was removed. They are `StorageError.ServerError.code`, `StorageError.Code` with its `init` and `rawValue`, and the 60 constants.

I did not run `./scripts/test-docs.sh`. The only doc change is the new property's comment, which has an external link and no symbol links.

## Review first

Five constants are named after the value on the wire rather than the server's enum key, because the key never leaves the server: `invalidAccessKeyId`, `maximumCredentialsLimit`, `conflictException`, `notFoundException` and `vectorBucketNotEmpty`.


grdsdev:
Great, but let's go with a typed raw representable structure similar to the ErrorCode in the Auth package, then add the known codes as `static let` constants.

Thanks for exposing that.

AndroidPoet:
Done in f969c72e.

`code` is now `StorageError.Code`, shaped like `ErrorCode` in Auth (`Decodable, RawRepresentable, Sendable, Hashable`, `init(rawValue:)` plus the shorthand `init(_:)`), with the 60 codes Storage defines today as documented `static let`s. I took them from [`src/internal/errors/codes.ts`](https://github.com/supabase/storage/blob/778e5336c0ee683525d924ca01cef9c049bb17ad/src/internal/errors/codes.ts) (v1.79.2), not from the docs page, which lists 32 of them.

```swift
catch let error as StorageError where error.serverError?.code == .keyAlreadyExists {
```

Two calls I would like you to check:

- I nested it as `StorageError.Code`, next to `StorageError.Kind`, rather than adding a top-level `ErrorCode`, which would collide with Auth's for anyone who does `import Supabase`. Happy to switch to a top-level `StorageErrorCode` if you prefer.
- Five constants take the value on the wire rather than the server's enum key, since the key never leaves the server: `invalidAccessKeyId`, `maximumCredentialsLimit`, `conflictException`, `notFoundException` and `vectorBucketNotEmpty`.

`serverErrorKeepsACodeItDoesNotKnow` pins that an unrecognised code still decodes and keeps its `rawValue`, which is the reason this is a struct rather than an enum. The 63 new symbols are registered under `storage.errors.error_codes` in `sdk-compliance.yaml`.

```
$ swift test --skip IntegrationTests   # exit 0, 1422 tests in 141 suites, 177 in StorageTests
$ swift format lint --strict --recursive Sources Tests Package.swift Examples   # exit 0
$ ./scripts/spell-check.sh   # exit 0, 413 files checked, 0 issues
```

The integration expectations moved to the constants too. They compile, but my last live run against Storage v1.72.1 predates this change, so I have not re-run them.

grdsdev (approved):

PR #1372, opened 2026-09-18
feat(postgrest)!: name a relation's schema with a type

AndroidPoet:
Implements the typed schema from #1317, as agreed there. A relation names its schema with a type, so querying it through the wrong schema stops compiling, and `from(_:)` can finally read the schema off the relation. That last part is the bug #1317 was opened for: `PostgrestRelation.schema` existed but nothing read it, so a relation outside `public` was queried in `public`.

```swift
enum PrivateSchema: PostgrestSchema {
  static let name = "private"
}

@Table("secrets", schema: PrivateSchema.self)
struct Secret {
  @PrimaryKey var id: Int
}

try await supabase.from(Secret.self).select().execute()                             // queries private
try await supabase.schema(PrivateSchema.self).from(Secret.self).select().execute()  // valid
try await supabase.schema(PrivateSchema.self).from(Todo.self).select().execute()    // does not compile
```

The last line fails with:

```
error: instance method 'from' requires the types 'PrivateSchema' and 'PublicSchema' be equivalent
```

## Shape

- `PostgrestSchema`, named as you asked, with `PublicSchema` as the default. Both are the uninhabited marker shape the request phases already use.
- `PostgrestRelation` gains `associatedtype Schema: PostgrestSchema = PublicSchema`. The default keeps every hand-written conformance compiling.
- `schema` is no longer a requirement. It stays as an extension property returning `Schema.name`, so the type is the only source of truth.
- `@Table("secrets", schema: PrivateSchema.self)` emits `typealias Schema = PrivateSchema`. Without `schema:` it emits `typealias Schema = PostgREST.PublicSchema`, qualified so a generated `PublicSchema` in the app's module cannot capture it. A `schema:` argument that is not a `T.self` literal gets the macro's diagnostic.
- `SupabaseClient` gets the typed `from(_:)` and `schema(_:)`, forwarding to its database client.
- The typed `schema(_:)` traps on a client that already has a schema, since choosing a second one is a programmer error.

## Behavior changes worth a look

- `from(Secret.self)` now sends `Accept-Profile: private`, taken from the relation. It previously sent nothing and queried `public`. A schema already set on the client wins, `"public"` included, and so does `SupabaseClientOptions(db: .init(schema:))`.
- A hand-written conformance that still declares `static let schema = "private"` compiles, but the string is no longer read. `V3_MIGRATION.md` says to declare `typealias Schema` instead.
- `@Table("secrets", schema: "private")` stops compiling. The argument fails type checking against `any PostgrestSchema.Type` before the macro expands, so the error is "cannot convert value of type 'String' to expected argument type 'any PostgrestSchema.Type'".

## Testing

- `PostgrestSchemaScopeTests`: the relation's schema on select and insert (`Accept-Profile` and `Content-Profile`), no profile for a public relation, the typed scope including `PublicSchema`, the client's schema winning including an explicit `"public"`, and the trap when scoping a client that already has a schema (exit test).
- `SupabaseClientTypedSchemaTests`: `supabase.from(Secret.self)` and `supabase.schema(PrivateSchema.self)` send `private`, and `db: .init(schema: "public")` wins.
- `TableIntegrationTests` compiles every `@Table` in its module next to a module-level `enum PublicSchema {}`, the shape postgres-meta's Swift generator emits.
- `DiagnosticsTests` pins the diagnostic for a `schema:` argument that is not a `T.self` literal.

```
$ swift test --skip IntegrationTests   # exit 0, 1435 tests
$ swift test --filter PostgrestMacrosTests.TableIntegrationTests   # exit 0, 31 tests
$ swift format lint --strict --recursive Sources Tests Package.swift Examples   # exit 0
$ ./scripts/spell-check.sh   # exit 0
$ ./scripts/test-docs.sh   # exit 0
```

The compliance job's validate, symbol and drift steps pass locally at the pinned tooling commit, and the new public symbols are registered under `database.query.schema_selection`.


grdsdev (changes requested):
Thanks, this is the shape we agreed on in #1317 and the compile error is the right one. A few things before it can land, then two design calls.

## Blocking

**1. The #1317 bug is still live for `SupabaseClient` users.** `SupabaseClient` has no typed `schema(_:)` or `from(_:)`, so the only route into the typed API is `client.schema("public").from(Secret.self)`. That sets `configuration.schema = "public"`, so `from<R>` takes the client-wins branch (`Sources/PostgREST/Query/PostgrestTypedSource.swift:23`) and sends `Accept-Profile: public`. Same silent misroute the issue was opened for. Please add the typed `schema(_:)` and `from(_:)` overloads to `SupabaseClient` in this PR, with a test that `supabase.from(Secret.self)` sends `Accept-Profile: private`.

**2. `schema` (String) and `Schema` (type) can disagree.** `from(_:)` routes by `R.schema`; the scope routes by `R.Schema`. A hand-written relation with `static let schema = "private"` and no `typealias Schema` compiles inside `schema(PublicSchema.self)` and queries `public`. The migration guide lists this as a caveat, but a query that is silently wrong must not compile. Drop `static var schema` from the protocol's requirement list and keep only the extension property, so it cannot be set apart from `Schema`. That also removes the migration bullet.

**3. `PublicSchema` collides with generated code.** The postgres-meta Swift template emits a `PublicSchema` namespace enum today. The macro emits an unqualified `typealias Schema = PublicSchema` into the user's module, so anyone with generated types gets their own enum, which does not conform, and every `@Table` stops compiling. Emit it qualified as `PostgREST.PublicSchema`, and add a macro test with a same-named local type.

## Should fix

**4. Macro accepts any member access.** `TableMacro.swift:41` takes `.base` without checking the member is `self`, so `schema: Schemas.privateOne` expands to `typealias Schema = Schemas`. Check for `.self` and diagnose otherwise.

**5. Write path untested.** #1317 covered inserts too. The new tests assert `Accept-Profile` only; please add one that an insert on `Secret` sends `Content-Profile: private`.

**6. Precedence tests.** Two cases with no test: `schema(PublicSchema.self).from(Todo.self)` sends `Accept-Profile: public` while `from(Todo.self)` sends nothing, and a client built with `schema: "public"` explicitly suppresses relation routing. Pin whichever behaviour you keep, and mention the explicit `"public"` case in the migration note.

## Nits

- Both new files are missing the file header block (AGENTS.md, "File Headers").
- `Legacy/PostgrestClient.swift:46` disambiguates the old `schema(_:)` in the Topics list but never adds the new overload, so it drops out of the DocC sidebar.
- `TableMacro.swift:43`: the `?? "PublicSchema"` fallback is unreachable since `expansion` already rejects a non-member-access argument. One accessor for the schema argument would do.
- `TableMacro.swift:39` and `DiagnosticsTests.swift:63`: the comments restate the code and narrate history; the PR description already carries that.

## Two design calls

**`client.schema("a").schema(PrivateSchema.self)` should not be allowed.** Once a schema is chosen, a second one is a programmer error. Add `precondition(configuration.schema == nil)` to the typed `schema(_:)`. That is why item 1 has to land in this PR: with the precondition, `client.schema("public").schema(PrivateSchema.self)` traps, so `SupabaseClient` needs the typed overloads for the typed scope to be reachable at all.

**Once in a typed scope, stay typed.** `client.schema(PublicSchema.self).from("table")` should not compile. Please drop `PostgrestSchemaScope.from(_ relation: String)` and its test from this PR. Shipping it now and removing it later would be a breaking change; leaving it out costs nothing. An explicit `untyped()` (or similar) to step back to the string API can come in a separate PR.

## Follow-up, not for this PR

The design docs still describe `schema` as a `String`: `docs/design/postgrest-v3.md:303`, `postgrest-v3-stage-1-plan.md:1319`, and the stage 5 generator template (`postgrest-v3-stage-5-plan.md:95`). I will update them once this lands.

AndroidPoet:
Thanks, all addressed in 8abbd35f. I checked each point against the code before changing anything, and every one held up.

**Blocking**

1. `SupabaseClient` now has the typed `from(_:)` and `schema(_:)`, forwarding to `rest`. `SupabaseClientTypedSchemaTests` checks that `supabase.from(Secret.self)` sends `Accept-Profile: private`, and covers the scope too.
2. The requirement is gone, and so is the migration bullet about the string and the type disagreeing. Dropping it left `static let schema = "public"` in 17 fixtures as dead code that still looked like it did something, so those are removed too.
3. The expansion emits `PostgREST.PublicSchema`. The collision test is compile-level rather than an expansion string: `TableIntegrationTests` declares a module-level `enum PublicSchema {}`, so every `@Table` in that module compiles against it.

**Should fix**

4. `schema:` must be `T.self`, and anything else gets the diagnostic. It is pinned with `schema: Schemas.privateOne`, which used to expand to `typealias Schema = Schemas` and fail with a conformance error.
5. Added `insertSendsTheRelationsProfile` (`Content-Profile: private`).
6. Added `thePublicScopeSendsItsProfile` and `anExplicitPublicSchemaOnTheClientWins`, and the migration note covers the second.

**Nits**: headers added. `PostgrestClient`'s Topics lists the typed `schema(_:)`, and `SupabaseClient`'s now disambiguates both pairs (DocC wants `->PostgrestSchemaScope<S>`, generic parameter included). One accessor with no fallback, and both comments removed.

**Design calls**: the precondition is in, covered by an exit test. `PostgrestSchemaScope.from(_: String)` and its test are gone.

A few things from my own pass:

- The PR body said `@Table("secrets", schema: "private")` gets the macro's diagnostic. It does not. The argument fails type checking against `any PostgrestSchema.Type` first, so the user sees "cannot convert value of type 'String'" and the macro never expands. The old test covered a path users cannot reach, so it is now the `Schemas.privateOne` case. Body corrected.
- With `SupabaseClientOptions(db: .init(schema: "public"))` the client's schema wins, so `supabase.from(Secret.self)` queries `public` and `supabase.schema(PrivateSchema.self)` hits the precondition. Both are in the migration note.
- A hand-written conformance that still declares `static let schema = "private"` compiles, but the string is no longer read (`from(_:)` did not read it before either). The migration note says to declare `typealias Schema` instead.

The design docs are yours, as you said.

